PR #1, opened 2019-07-25
Create project repo initial setup

jackbsteinberg:


.eslintrc.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  'env': {
+    'commonjs': true,
```
domenic: Looks like this is redundant with `'node': true`

.gitignore
```diff
@@ -0,0 +1 @@
+/node_modules
```
domenic: Nit: newline at end of file

README.md
```diff
@@ -0,0 +1 @@
+# LAPI Rewriter Script
```
domenic: Get-originals Rewriter Codemod?

package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "lapi-rewriter",
+  "version": "1.0.0",
+  "description": "Rewriter presubmit script for LAPIs",
```
domenic: Name and description update to focus on get-originals.

package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "lapi-rewriter",
+  "version": "1.0.0",
```
domenic: I prefer 1.0.0-pre until you actually do `npm publish`

package.json
```diff
@@ -0,0 +1,23 @@
...
+  "description": "Rewriter presubmit script for LAPIs",
+  "main": "lib/index.js",
+  "directories": {
```
domenic: `directories` doesn't actually do anything, so best to remove it

package.json
```diff
@@ -0,0 +1,23 @@
...
+    "lib": "lib"
+  },
+  "dependencies": {
+    "jest": "^24.8.0",
```
domenic: I think jest is a dev dependency?

package.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+  "devDependencies": {
+    "eslint": "^6.1.0",
+    "eslint-config-google": "^0.13.0"
+  },
+  "scripts": {
+    "start": "node lib/index.js",
+    "test": "jest"
+  },
+  "author": "jacksteinberg@chromium.org",
```
domenic:
```suggestion
  "author": "Jack Steinberg <jacksteinberg@chromium.org>",
```

.eslintrc.js
jackbsteinberg: This file was autogenerated (I picked Google's JS standard, which I assume we theoretically adhere to), I was inclined not to mess with it but I can remove this if you want.

package.json
jackbsteinberg: This was also autogenerated, I didn't install it as --save-dev though, so maybe I should go back and do that.

package.json
```diff
@@ -0,0 +1,20 @@
...
+    "eslint": "^6.1.0",
+    "eslint-config-google": "^0.13.0",
+    "jest": "^24.8.0"
+  },
+  "scripts": {
+    "start": "node lib/index.js",
+    "test": "jest"
+  },
+  "author": "Jack Steinberg <jacksteinberg@chromium.org>",
+  "repository": "https://github.com/jackbsteinberg/lapi-rewriter.git"
```
domenic:
```suggestion
  "repository": "https://github.com/jackbsteinberg/get-originals-rewriter.git"
```

package.json
```diff
@@ -0,0 +1,20 @@
...
+  "dependencies": {
+    "jscodeshift": "^0.6.4"
+  },
+  "devDependencies": {
+    "eslint": "^6.1.0",
+    "eslint-config-google": "^0.13.0",
+    "jest": "^24.8.0"
+  },
+  "scripts": {
+    "start": "node lib/index.js",
```
domenic: Looks like this isn't useful either; best to remove

jackbsteinberg:
> Nooo the test script is good, just the start one

Oh, I figured if we don't need `npm start` to run the rewriter itself why would we need `npm test` to run the tests when just writing `jest` will do the trick

domenic:
`npm test` is the standard way of running the tests, and is generally applicable, but `npm start` is more meant for servers. Having `npm start` doesn't buy us much here because people won't use this by cloning the repo then doing `npm start`. Instead they'd use it more like https://github.com/jongold/import-codemod#usage (presumably without the config file).

domenic:
When merging, don't forget to squash and merge. (You may want to use the repository settings to turn off the other possibilities, or at least the "Create a merge commit" option.)

.eslintrc.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
fergald: Is there a reason we're not using the same eslintrc as the other projects?

PR #2, opened 2019-07-25
Write console method removal script

jackbsteinberg:
Exploring AST codemods with jscodeshift by writing functions to strip console statements, with tests


domenic:
Oh. It'd be a good idea to also update the README with basic usage and explanation of what the repo does as we add new functionality. E.g. with small examples. This is generally preferable to doing a big documentation sprint all at once.

lib/remove-console-statements.js
```diff
@@ -0,0 +1,39 @@
+const cleanConsoleLogs = (j, root) => {
```
fergald: Please document functions/methods with jsdoc. For switch, almost everything was an impl of a spec, so didn't need inline docs but here, we need proper docs.

lib/remove-console-statements.js
```diff
@@ -0,0 +1,39 @@
...
+  root.find(j.CallExpression, {
```
fergald: Seems like this could all live in a function findConsoleCalls or findVariableCalls('console').

Independently I'm a bit concerned that

let console = whatever();
console.foo();

is going to get replaced. I'd like us to have a failing test with that in it, so we can fix it in the future.

What I found when using the previous eslint was that the only allowed external reference in an mjs file was to window (I'm sure there were others but I don't know what they were). I couldn't call requestAnimationFrame, I had to call window.requestAnimationFrame. So we can enforce that with lint and then say that actually this removes calls to window.console.foo and if you break that lint rule, you're on your own.

lib/__tests__/remove-console-statements.js
```diff
@@ -0,0 +1,3 @@
+jest.autoMockOff();
+const defineTest = require('jscodeshift/src/testUtils').defineTest;
+defineTest(__dirname, 'remove-console-statements');
```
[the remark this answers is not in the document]
jackbsteinberg: Yeah I can refactor it so it's like this, but if I want to unit test the different codemod functions (like getConsoleStatments and getConsoleExpressions) then I need to split them into different files.

I'll go ahead and do that now

lib/__testfixtures__/remove-console-statements.input.js
```diff
@@ -0,0 +1,8 @@
+const tips = ['a', 'b', 'c'];
```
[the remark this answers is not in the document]
jackbsteinberg: What I'm gonna do here is add tests for each, while at the same time keeping this test. That way both the console removal units are tested, along with their interaction.

lib/__tests__/remove-console-statements.js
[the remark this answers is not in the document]
jackbsteinberg: So I see my confusion with this - because of the way I named it it seems like you would have to create a new file for each test, but if you use it like this:

```js
// test.js
defineTest(__dirname, 'rewriter', undefined, 'remove-console-all');
defineTest(__dirname, 'rewriter', undefined, 'remove-console-all-2');
```

Then you can test multiple fixtures for one file. Learned this while I was refactoring to allow for tests, so now you can unit test individual mods I guess 🎉

lib/remove-console-statements.js
jackbsteinberg: This is something we may need to figure out going forward, but there is something to be said about this script only being run on code we write, so we could just enforce a standard in our own development that you shouldn't do things like reassign `console`

domenic: Yeah, I think supporting this kind of uncooperative code is a "v2" feature. If anything we could require more explicit author intent, e.g. comments to opt-out of transformation in such cases.

domenic:
So unfortunately I think I miscommunicated some of the advice here. I don't think we need to have a separate unit test framework. And we don't need to have separate mods for expressions and statements; we should still just have one console-remover mod. (Maybe just one big rewriter mod.)

So in particular I think the console mod can be in just one file, not four. And I think the test file should look something like

```js
'use strict';
jest.autoMockOff();
const defineTest = require('jscodeshift/src/testUtils').defineTest;

defineTest(__dirname, 'rewriter', undefined, 'remove-console-expressions');
defineTest(__dirname, 'rewriter', undefined, 'remove-console-statements');
defineTest(__dirname, 'rewriter', undefined, 'remove-console-all');
```

or similar, without the testrunner.js/defineUnitTest.

---

Ideally, I'd like to have the test file look like

```js
'use strict';
jest.autoMockOff();
const defineTest = require('jscodeshift/src/testUtils').defineTest;

runAllTests();
```

so that we don't need to add a new line every time you add input/output files. I thought that was something jscodeshift had built in. But upon reviewing examples, I don't think it does. So let's leave making that work for a follow-up.

---

Also, you're going to need to rebase this on master, to fix the issue where the diff still shows the files that are already present (like package.json).

lib/remove-console-statements.js
fergald: I think there must be an easy-ish way to do this because eslint can tell the difference between

```js
console.log("foo");
```

and

```js
const console = null;
console.log("foo");
```

lib/__tests__/test.js
```diff
@@ -0,0 +1,4 @@
+jest.autoMockOff();
+const runTestFixtures = require('../runTestFixtures.js');
```
domenic: Very nice! However I think there's not much point in separating this into a runTestFixtures.js file. You can just keep the code here; after all, that code is entirely concerned with running the tests.

lib/__tests__/test.js
```diff
@@ -0,0 +1,4 @@
+jest.autoMockOff();
```
domenic: You should put `"use strict";` as the first line of all code files. Add the ESLint rule `strict: ["error", "global"]` to enforce this.

lib/mods/clean-console.js
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param {any} j - the jscodemod api
+ * @param {Collection} root - the root Collection
+ */
+const cleanConsole = (j, root) => {
+  cleanConsoleExpressions(j, root);
+  cleanConsoleStatements(j, root);
+};
+
+module.exports = cleanConsole;
```
domenic: This and lib/rewriter.js are inconsistent. You should pick either `module.exports = functionName` style or `module.exports = () => { ... }` style.

lib/mods/utils.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @param {any} j - the jscodemod api
+ * @param {Collection} root - the root Collection
+ * @return {Collection} - a Collection of console statements
+ *
+ * @example
+ * getConsoleStatments(j, root).filter(...);
+ */
+
+const getConsoleStatments = (j, root) => {
```
domenic: It's a style choice but you could also do `exports.getConsoleStatements = (j, root) => { ... }` instead of separately assigning `module.exports`.

lib/runTestFixtures.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
+path = require('path');
```
domenic: Missing `const`. Add the ESLint rule https://eslint.org/docs/rules/no-implicit-globals to fix this.

Maybe just port the .eslintrc.js from our other projects, as Fergal suggested?

lib/runTestFixtures.js
```diff
@@ -0,0 +1,33 @@
...
+const defineTest = require('jscodeshift/src/testUtils').defineTest;
+
+/**
+ * Runs all testfixtures in folder with rewriter script
+ */
+function runTestFixtures() {
```
domenic: Your style here seems to be inconsistent with elsewhere, where you chose `const x = () => {}` for all your functions. https://eslint.org/docs/rules/func-style can enforce one or the other.

lib/mods/utils.js
jackbsteinberg: I think I'm going to go the route of separately defining all the functions before exporting them, and I like the format where they all export as an object:

```js
module.exports = {
  getConsoleStatements,
  ...
}
```

lib/runTestFixtures.js
jackbsteinberg: So the other projects use babel-eslint, where this project is using CommonJS per the Google doc. I could port the .eslintrc.js from the other projects, but that would also require going in and changing all the CommonJS import / exports

lib/__tests__/test.js
fergald: Can we just use .mjs as much as possible?

domenic: Not in Node.js, no.

lib/mods/clean-console.js
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Replaces console statements without ExpressionStatement parents
+ * with `undefined`.
+ *
+ * @param {any} j - the jscodemod api
+ * @param {Collection} root - the root Collection
+ *
+ * @return {Collection} - the modified root Collection
+ *
+ * @example
```
fergald: This example does not seem useful, it's exactly the signature of the method. Similar below.

lib/mods/utils.js
```diff
@@ -0,0 +1,25 @@
...
+ * Gets the Collection of CallExpressions with callee name console.
+ *
+ * @param {any} j - the jscodemod api
+ * @param {Collection} root - the root Collection
+ * @return {Collection} - a Collection of console statements
+ *
+ * @example
+ * getConsoleStatements(j, root).filter(...);
+ */
+function getConsoleStatements(j, root) {
```
fergald: why is this in a separate file when there is only one file that uses it? It seems like it could be in that file.

lib/__tests__/test.js
```diff
@@ -0,0 +1,36 @@
...
+
+const fs = require('fs');
+const path = require('path');
+const defineTest = require('jscodeshift/src/testUtils').defineTest;
+
+/**
+ * Runs all testfixtures in folder with rewriter script
+ */
+function runTestFixtures() {
+  walkDir(path.join(__dirname, '../__testfixtures__'), (file) => {
```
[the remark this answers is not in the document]
jackbsteinberg: That's how jscodemod told me to set up the file structure

```
jscodeshift comes with a simple utility to allow easy unit testing with Jest, without having to write a lot of boilerplate code. This utility makes some assumptions in order to reduce the amount of configuration required:

The test is located in a subdirectory under the directory the transform itself is located in (eg. __tests__)
Test fixtures are located in a __testfixtures__ directory
This results in a directory structure like this:

/MyTransform.js
/__tests__/MyTransform-test.js
/__testfixtures__/MyTransform.input.js
/__testfixtures__/MyTransform.output.js
```

.eslintrc.js
```diff
@@ -14,5 +14,12 @@ module.exports = {
     'ecmaVersion': 2018,
   },
   'rules': {
+    'func-style': [
+      "error",
+      "declaration",
```
domenic: I think you want "expression" based on what I've seen of your style so far. (Then allowArrowFunctions is not necessary).

.eslintrc.js
```diff
@@ -14,5 +14,12 @@ module.exports = {
...
+      { "allowArrowFunctions": true }
+    ],
+    'no-implicit-globals': 'error',
+    'strict': ["error", "global"]
```
domenic: Nit: use a single quote style in this file, either single or double, not both.

.eslintrc.js
domenic: Sorry, I see you switched everything to declarations. In that case maybe try removing allowArrowFunctions?

jackbsteinberg: Do I need to completely disallow arrow functions?

domenic: If I am reading the docs right, removing "allowArrowFunctions" will not totally disallow arrow functions. It will just prevent you from using them when declaring new functions. Which is good, because you've decided to stick to `function foo() { }`, instead of `const foo = () => { }`.

PR #5, opened 2019-07-30
Add rewriter script and initial explainer text to go with it

jackbsteinberg:


README.md
```diff
@@ -1,3 +1,54 @@
...
+
+Applies rewriter logic recursively to every PATH.
+If --stdin is set, each line of the standard input is used as a path.
+
+Options:
+"..." behind an option means that it can be supplied multiple times.
+All options are also passed to the transformer, which means you can supply custom options that are not listed here.
+
+      --(no-)babel              apply babeljs to the transform file
+                                (default: true)
```
domenic: This seems bad... filed #6.

README.md
```diff
@@ -1,3 +1,54 @@
...
+Usage: npm run rewrite PATH...
+  or:  npm run rewrite -- [OPTION]... PATH...
+  or:  npm run rewrite -- [OPTION]... --stdin < file_list.txt
+
+Applies rewriter logic recursively to every PATH.
+If --stdin is set, each line of the standard input is used as a path.
+
+Options:
+"..." behind an option means that it can be supplied multiple times.
+All options are also passed to the transformer, which means you can supply custom options that are not listed here.
```
domenic: Let's remove this sentence since we don't have any custom options.

README.md
```diff
@@ -1,3 +1,54 @@
...
+
+      --(no-)babel              apply babeljs to the transform file
+                                (default: true)
+  -c, --cpus=N                  start at most N child processes to process source files
+                                (default: max(all - 1, 1))
+  -d, --(no-)dry                dry run (no changes are made to files)
+                                (default: false)
+      --extensions=EXT          transform files with these file extensions (comma separated list)
+                                (default: js)
```
domenic: Seems bad, filed #7.

README.md
```diff
@@ -1,3 +1,54 @@
...
+      --(no-)run-in-band        run serially in the current process
+                                (default: false)
+  -s, --(no-)silent             do not write to stdout or stderr
+                                (default: false)
+      --(no-)stdin              read file/directory list from stdin
+                                (default: false)
+  -v, --verbose=0|1|2           show more information about the transform process
+                                (default: 0)
+      --version                 print version and exit
+```
```
domenic: Let's remove docs for --parser, --parser-config, and --babel. Changing those is not a supported scenario for this rewriter.

README.md
jackbsteinberg: Addressing this in next commit

README.md
jackbsteinberg: Addressing this in next commit

README.md
```diff
@@ -1,3 +1,46 @@
...
+
+```
+Usage: npm run rewrite PATH...
+  or:  npm run rewrite -- [OPTION]... PATH...
+  or:  npm run rewrite -- [OPTION]... --stdin < file_list.txt
+
+Applies rewriter logic recursively to every PATH.
+If --stdin is set, each line of the standard input is used as a path.
+
+Options:
```
fergald: Is this just the output of --help? If so, I don't think it's a good idea to include it in the README as it will need updating every time it changes. Better to just explain how get it to dump out the current set of options.

Poking a bit more, it seems like doing `npm run rewrite --help` gets you nothing but adding script entry of

```
  "rewrite-help": "jscodeshift -h"
```

works, so maybe point people at that.

domenic: How do you anticipate it changing?

Per the README you need -- OPTIONS, e.g. -- --help.

fergald: Ah `npm run rewrite -- --help` does the right thing.

> How do you anticipate it changing?

How do you anticipate it never changing? Unless I'm missing a reason why this is all set in stone, it seems highly unlikely that the flags and their help text will never change. All of that content (flag names, values, help-text) is already checked in, the principle of "don't repeat yourself" says we we shouldn't check it in again in a new location.

domenic: It would change if we bump the jscodeshift version, which seems rare (very few new versions recently).

I think there's generally a tradeoff between "don't repeat yourself" and "have good documentation"; after all documentation is often just a repetition of what you can glean from the source code. My personal opinion is that any CLI tool on GitHub should state how you use and configure it in the README, instead of requiring you to clone the project, install its dependencies, and then run its `--help` to find that information.

fergald:
Fair enough about not having to install it. It would be nice though if there was a simple way to regenerate the docs, e.g. putting this in USAGE.txt so that we can have a script entry of

```
docs: "jscodeshift --help > USAGE.txt"
```

(jscodeshift has a `docs` entry, although they don't update their README.md as part of it).

But I'm also happy to wait until this changes before doing anything to make keeping up with changes easier.

PR #9, opened 2019-07-30
Update README to reflect current rewriter capabilities

jackbsteinberg:


README.md
```diff
@@ -2,6 +2,16 @@
 
 [![Build Status](https://travis-ci.com/jackbsteinberg/get-originals-rewriter.svg?branch=master)](https://travis-ci.com/jackbsteinberg/get-originals-rewriter)
 
+## Introduction
+
+When adding built-in modules to the web platform,
```
domenic: Let's note how this is a Chromium-specific concern. Something like "Chromium is implementing some built-in module proposals in unprivileged JavaScript. To make this possible, we need..."

README.md
```diff
@@ -2,6 +2,16 @@
...
+we need a way to ensure they're accessing the "original versions"
+of the methods and properties of the global built-ins they rely on.
+We do so using the `std:global` built-in module described in the 
+[Get-Originals Proposal](https://github.com/domenic/get-originals) by [Domenic Denicola](https://github.com/domenic).
+This Get-Originals Rewriter script will run before Chromium code is submitted,
+to convert existing code to using "original versions" and prevent future code from checking in without them.
+
```
domenic: We could also note that this project could be generally useful to folks who want to use get-originals in their programs, for [other reasons](https://github.com/domenic/get-originals#motivations). Although that isn't our primary goal.

PR #10, opened 2019-07-30
Scaffold constructor replacement mod

jackbsteinberg:
I'm putting this PR out to get feedback on the general scaffolding of how the replacement of constructors would work, the feature itself requires more fleshing out.

What still needs to be added:
- `isGlobal` implementation
  - Will likely require pulling an updated list
     of globals that need originals
- `getOriginal` implementation
  - Same as above
- `addImport` implementation

### DIFF
--- a/lib/__testfixtures__/replace-constructors.input.js
+++ b/lib/__testfixtures__/replace-constructors.input.js
@@ -1,0 +1,3 @@
+const n = new Number();
+const t = new TypeError();
+const s = new Set([1, 2, 3]);
--- a/lib/__testfixtures__/replace-constructors.input.js
+++ b/lib/__testfixtures__/replace-constructors.input.js
@@ -1,3 +1,4 @@
 const n = new Number();
 const t = new TypeError();
 const s = new Set([1, 2, 3]);
+const u = new URIError('Error');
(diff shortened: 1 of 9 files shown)


lib/mods/replace-constructors.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+
+/**
+ * Checks if the given constructor name is global
+ *
+ * @param {String} constructor - the name of the constructor
+ *
+ * @return {Boolean} boolean indicating if the constructor is global
```
domenic:
```suggestion
 * @return {Boolean} whether the constructor is global
```

lib/mods/replace-constructors.js
```diff
@@ -27,15 +39,21 @@ function getGlobalNewExpressions(j, root) {
...
+function addImport(module, original, root) {
   // TODO(jackbsteinberg): implement
-  return 'original_' + name;
+  // if an import of original from module exists, noop
+
+  // if an import from module exists without importing original,
+  // add original to the import
+
+  // if no import from module exists, add a new import
+  // of original, from module
```
domenic: Be sure to add tests for all three of these cases. Probably involving multiple test files!

lib/__testfixtures__/replace-constructors.input.js
```diff
@@ -0,0 +1,3 @@
+const n = new Number();
```
domenic: Note that these are all JS globals. We should also test (probably best in a separate file?) some web globals.

PR #11, opened 2019-07-31
Write update-globals script to fetch Browser & JS globals

jackbsteinberg:
Unsure about the best way to manage this but right now this PR will merge into constructors, a branch with an open PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,5 +20,6 @@ module.exports = {
     ],
     'no-implicit-globals': 'error',
     'strict': ['error', 'global'],
+    'dot-notation': "error"
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,6 @@ module.exports = {
     ],
     'no-implicit-globals': 'error',
     'strict': ['error', 'global'],
-    'dot-notation': "error"
+    'dot-notation': 'error'
   },
 };
--- a/lib/update-globals.js
+++ b/lib/update-globals.js
@@ -1,0 +1,73 @@
+'use strict';
+
+const fs = require('fs');
+const vm = require('vm');
+const path = require('path');
+const https = require('https');
+
+// collect globals
+const jsGlobals = {};
+
+vm.createContext(jsGlobals);
+vm.runInContext(`
+  this.props = Object.getOwnPropertyNames(this)`, jsGlobals);
+
+const windowGlobals = new Promise((resolve, reject) => {
+  https.get('https://tidoust.github.io/reffy-reports/whatwg/crawl.json',
+      (response) => {
+        let data = [];
+
+        response.on('data', (d) => {
+          data += d;
+        });
+
+        response.on('end', () => {
+          resolve(JSON.parse(data));
+        });
+      }).on('error', (err) => {
+    console.error('could not get WHATWG globals');
+    reject(err);
+  });
+});
+
+// create a global.json file (overwriting if it exists)
+/**
+ * Combines the globals into a stringified Set
+ *
+ * @param {Array} arr - an array of global variables
+ *
+ * @return {String} a stringified set of globals
+ */
+function listGlobals(arr) {
+  let str = `new Set([`;
+  arr.forEach((elt) => {
+    str += elt + `,\n    `;
+  });
+  return str.substr(0, str.length-2) + `]);\n`;
+}
+
+windowGlobals.then((data) => {
+  // get data into a nice format
+  data.results.forEach((elt) => {
+    if (elt.idl['idlNames']) {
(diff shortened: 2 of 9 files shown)


lib/update-globals.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param {Array} arr - an array of global variables
+ *
+ * @return {String} a stringified set of globals
+ */
+function listGlobals(arr) {
+  let str = `new Set([`;
+  arr.forEach((elt) => {
+    str += `'${elt}',\n    `;
+  });
```
domenic: I think this might be simpler with `array.join()`, in particular, then you wouldn't need the `.substr()` at the end.

lib/update-globals.js
```diff
@@ -0,0 +1,74 @@
...
+
+// collect globals
+const jsGlobals = {};
+
+vm.createContext(jsGlobals);
+vm.runInContext(`
+  this.props = Object.getOwnPropertyNames(this)`, jsGlobals);
+
+const windowGlobals = new Promise((resolve, reject) => {
+  https.get('https://tidoust.github.io/reffy-reports/whatwg/crawl.json',
```
domenic: Sigh, Node.js's built-in HTTP library is so terrible. If we do this more than once, let's use a library. But I think this code is good for now; the cost of adding a library dependency is not worth cleaning it up.

lib/update-globals.js
```diff
@@ -0,0 +1,74 @@
+'use strict';
```
domenic: Let's put this in a `scripts/` directory. Then when publishing to npm (or, more likely in our case, importing into the Chromium tree) we can avoid including that directory.

The output should still be in `lib/` though.

lib/update-globals.js
```diff
@@ -0,0 +1,74 @@
...
+  // combine with jsGlobals
+  const combinedGlobals = props.concat(jsGlobals.props);
+
+  // write to file
+  const fileText = `'use strict';
+
+module.exports =
+  ${listGlobals(combinedGlobals)}`;
+
+  fs.writeFile(path.join(__dirname, 'global.js'), fileText, () => {
```
domenic: Nit: I might call this `globals-set.js` or similar.

lib/update-globals.js
jackbsteinberg: I added `scripts/*` to the `.gitignore`, I'm assuming we want that

lib/update-globals.js
jackbsteinberg: SGTM, it is a bit of a mess

lib/update-globals.js
jackbsteinberg: Oh I did this backwards, let me fix now

scripts/update-globals.js
```diff
@@ -0,0 +1,93 @@
...
+        }).on('error', (err) => {
+      console.error('could not get WHATWG globals');
+      reject(err);
+    });
+  });
+
+  // Take relevant globals data from crawl
+  let globals = [];
+  windowGlobals.results.forEach((elt) => {
+    if (elt.idl['idlNames']) {
```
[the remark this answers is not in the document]
domenic: Might want to add https://eslint.org/docs/rules/dot-notation

scripts/update-globals.js
```diff
@@ -0,0 +1,93 @@
...
+
+          response.on('data', (d) => {
+            data += d;
+          });
+
+          response.on('end', () => {
+            resolve(JSON.parse(data));
+          });
+        }).on('error', (err) => {
+      console.error('could not get WHATWG globals');
```
domenic: This will be taken care of by the console.error on main, so just rejecting is fine

scripts/update-globals.js
```diff
@@ -0,0 +1,93 @@
...
+    });
+  });
+
+  // Take relevant globals data from crawl
+  let globals = [];
+  windowGlobals.results.forEach((elt) => {
+    if (elt.idl['idlNames']) {
+      // eslint-disable-next-line no-unused-vars
+      const {_dependencies, _reallyDependsOnWindow, ...globalNames} =
+          elt.idl['idlNames'];
```
domenic: It looks like this will also catch things that are just IDL constructs, like enums.

I think filtering for `type` of "namespace" or "interface" is the best bet here.

jackbsteinberg: Added a filter to this and wow do we have fewer entries now (~60 now vs ~1500 before)

.eslintrc.js
```diff
@@ -20,5 +20,6 @@ module.exports = {
     ],
     'no-implicit-globals': 'error',
     'strict': ['error', 'global'],
+    'dot-notation': "error"
```
domenic:
```suggestion
    'dot-notation': 'error'
```

Issue #17, opened 2019-08-05
Format for different types of imports


jackbsteinberg:
The imports for Get-Originals break down into about 4 categories, each with a different syntax:

* Constructors: `import TypeError from 'std:global/TypeError'`
  * Imported as a default import, with their name matching the global version
* Static Methods: `import { parseInt_static } from 'std:global/Number'`
  * Imported as a named import under `name+_static`
* Non-Static Methods: `import { toExponential as Number_toExponential } from 'std:global/Number'`
  * Imported as a named import under `name`
* Getters / Setter: `import { returnValue_get, returnValue_set } from 'std:global/Event'`
  * Imported as a name import under `name_get/set`

[Source](https://github.com/domenic/get-originals#the-api)

PR #18, opened 2019-08-05
Add imports to file for replaced constructors

jackbsteinberg:
Not all cases (described in #17) are covered yet, but this works for replacing constructors


lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {string} mod - the module of the identifier
+ *
+ * @return {string} the name of the original version
+ */
+function getOriginal(name, type, mod) {
+  switch (type) {
+    case 'constructor':
+      return name;
+      break;
+    case 'static':
```
domenic: Let's not include stuff that isn't tested. Instead just do constructor + `default: throw new Error('not yet implemented')`.

lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+ * @param {string} type - the type of namespace of the identifier
+ * @param {string} mod - the module of the identifier
+ *
+ * @return {string} the name of the original version
+ */
+function getOriginal(name, type, mod) {
+  switch (type) {
+    case 'constructor':
+      return name;
+      break;
```
domenic: Nit: No need for `break` after `return`

lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+ */
+function isGlobal(node) {
+  return nodeInTSDefs(node) || nodeInIDLSpecs(node);
+}
+
+/**
+ * Gets the original name of the identifier param
+ *
+ * @param {string} name - the name of the identifier
+ * @param {string} type - the type of namespace of the identifier
```
domenic: I don't think it's a type of namespace. Maybe "type of original"?

lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+  for (const module in IMPORTS) { // eslint-disable-line guard-for-in
+    sourceFile.addImportDeclaration({
+      defaultImport: IMPORTS[module][0],
+      moduleSpecifier: `std:global/${module}`,
+    });
+  }
+
+  IMPORTS = {};
+
+  /*
```
domenic: Seems leftover; best removed before checkin.

lib/mods/replace-constructors.js
```diff
@@ -71,8 +16,10 @@ function replaceGlobalConstructors(node) {
   if (nodeTypes.isNewExpression(node) &&
       isGlobal(node)) {
     const identifier = node.getExpression();
-    identifier.replaceWithText(getOriginal(
-        node.getType().getSymbol().getName()));
+    const mod = node.getType().getSymbol().getName();
+    const original = getOriginal(mod, 'constructor');
+    identifier.replaceWithText(original);
+    updateImports(mod, original);
```
domenic: The use of hidden global state, i.e. `IMPORTS` inside the helpers.js file, is not a great pattern here. However, I can understand the motivation; it would be annoying to have to add the imports as you go, because then you'd have to see if any were previously added.

Instead, what about an architecture where each mod (e.g. `replaceGlobalConstructors`) accepts and updates an `imports` object. Then you create that object inside `rewriter.js`, pass it to each of the mods, and then pass it to `addAllImports` in the end.

lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+    defaultImport?: string;
+    namespaceImport?: string;
+    namedImports?: (OptionalKind<ImportSpecifierStructure> |
+       string | WriterFunction)[] | WriterFunction;
+    moduleSpecifier: string;
+  */
+}
+
+/**
```
domenic: Let's leave all these functions where they are, until we actually are using them in two source files. I suspect they will change when they do.

Then this file can just be something like `imports.js` or `imports-helpers.js` instead of a grab-bag full of all sorts of stuff.

lib/helpers.js
```diff
@@ -0,0 +1,121 @@
...
+  IMPORTS[mod] = [original];
+}
+
+/**
+ * Adds all imports from the list to the top of the source file
+ *
+ * @param {SourceFile} sourceFile - the file to add the imports to
+ */
+function addAllImports(sourceFile) {
+  for (const module in IMPORTS) { // eslint-disable-line guard-for-in
```
domenic: You can do `for (const [module, import] of Object.entries(IMPORTS)) { ... }`, and avoid the ESLint warning.

lib/__testfixtures__/replace-constructors-web.input.js
```diff
@@ -1,2 +1,2 @@
-const de = new DOMException('Error');
+const de = new DOMException("Error");
```
domenic: Probably best to leave this as-is, so that it's clear that this PR only changes the output of the tests.

lib/helpers.js
jackbsteinberg: Are you referring to the `IMPORTS = {};` or the comment?

domenic: The comment.

lib/mods/replace-constructors.js
jackbsteinberg: Yeah I thought a bit about what the best pattern for this would be, I agree that hidden global state should be cautioned against (and didn't think of this way to do it at the time). I'll switch to this.

lib/helpers.js
jackbsteinberg: Hmm, I do favor changing the name to something more descriptive, but to me it doesn't really feel like a 'grab-bag', all these functions to me are critical for the core of handling the originals, checking, grabbing, and importing. I see the benefit of renaming, but I don't see the benefit of putting them back where they were.

domenic: I think checking, grabbing, and importing are all very different. Putting them in one file is what makes it feel like a grab-bag.

In addition, all the non-imports functions right now are rather constructor-specific, so they should stay there until/unless they get generalized.

lib/__testfixtures__/replace-constructors-web.input.js
jackbsteinberg: I was getting an ESLint warning for this

domenic: We should disable ESLint on `__testfixtures__`

jackbsteinberg: Oh that's strange, I did, I'm not sure why it was complaining then

lib/mods/replace-constructors.js
```diff
@@ -3,6 +3,30 @@
...
+ */
+function updateImports(mod, original, imports) {
+  if (mod in imports) {
+    if (original in imports[mod]) {
+      // if an import of original from module exists, noop
+      return;
+    }
+    // if an import from module exists without importing original,
+    // add original to the import
+    imports[mod].append(original);
```
domenic: There is no append() method on arrays. The fact that this slipped through means the tests are not covering this.

If this code path isn't testable currently, then, we should remove it until we add some tests that exercise it.

lib/mods/replace-constructors.js
```diff
@@ -3,6 +3,30 @@
...
+/**
+ * Adds an import for the module to the list
+ *
+ * @param {string} mod - the name of the module to import from
+ * @param {string} original - the name of the original to get
+ * @param {Object} imports - the current list of imports
+ */
+function updateImports(mod, original, imports) {
+  if (mod in imports) {
+    if (original in imports[mod]) {
```
domenic: Is this code tested? It seems like it could be via a test file that does `new Something()` twice.

lib/mods/replace-constructors.js
jackbsteinberg: Yup, `.append()` is python, good catch. I'll write up a test for these cases

jackbsteinberg: Hmm, currently the code only gets the originals for constructors, meaning this case won't come up until a mod for method calls or statics is written.

domenic: Best to delete any non-used code for now, then. I suspect it will change in the future.

PR #20, opened 2019-08-07
Add method call replacement mod and reorganized import / originals

jackbsteinberg:
Created a method call replacement mod and pulled import and originals functionality into their own files

### DIFF
--- a/lib/__testfixtures__/property-access.output.js
+++ b/lib/__testfixtures__/property-access.output.js
@@ -1,0 +1,5 @@
+import { call as Reflect_call } from "std:global/Reflect";
+import { toString as Array_toString } from "std:global/Array";
+const a = [1, 2, 3];
+const l = a.length;
+const str = Reflect_call(a, Array_toString, []);
--- a/lib/__testfixtures__/property-access.output.js
+++ b/lib/__testfixtures__/property-access.output.js
@@ -1,5 +1,5 @@
-import { call as Reflect_call } from "std:global/Reflect";
 import { toString as Array_toString } from "std:global/Array";
+import { call as Reflect_call } from "std:global/Reflect";
 const a = [1, 2, 3];
 const l = a.length;
 const str = Reflect_call(a, Array_toString, []);
--- a/lib/imports.js
+++ b/lib/imports.js
@@ -37,7 +37,7 @@ function updateImports(mod, identifier, type, imports) {
   // TODO(jackbsteinberg): update this for other cases
   // add original for `call` from `Reflect` upon first method call
   if (type === 'method' && !imports.Reflect) {
-    imports.Reflect = {IMPORTS: ['call']};
+    imports.Reflect = {IMPORTS: ['call as Reflect_call']};
   }
 
   if (mod in imports) {
--- a/lib/imports.js
+++ b/lib/imports.js
@@ -34,41 +34,35 @@ function generateImportText(identifier, type, mod) {
 function updateImports(imports, mod, identifier, type) {
   const importText = generateImportText(identifier, type, mod);
 
-  if (type === 'method' && !imports.Reflect) {
-    imports.Reflect = {namedImports: ['call as Reflect_call']};
-  } else if (imports.Reflect) {
-    // if Reflect in imports namedImports will be an initialized array with
-    // values, as Reflect can't be constructed and won't be default imported
-    imports.Reflect.namedImports.push('call as Reflect_call');
-  }
-
-  if (mod in imports) {
-    if (type === 'constructor') {
-      imports[mod].defaultImport = importText;
-      return;
-    }
-
-    if (imports[mod].namedImports.includes(identifier)) {
-      // if an import of original from module exists, noop
-      return;
-    }
-    // if an import from module exists without containing the identifier,
-    // add it to the import list
-    imports[mod].namedImports.push(importText);
-    return;
-  }
-
-  if (type === 'constructor') {
+  if (!(mod in imports)) {
     imports[mod] = {
-      defaultImport: importText,
       namedImports: [],
     };
-    return;
   }
 
-  // if no import from module exists, add a new import
-  // of identifier, from module
-  imports[mod] = {namedImports: [importText]};
+  switch (type) {
+    case 'constructor':
+      imports[mod].defaultImport = importText;
+      return;
+    case 'method':
+      if (!imports.Reflect) {
+        imports.Reflect = {namedImports: ['call as Reflect_call']};
+      } else if (!imports.Reflect.namedImports.includes('call as Reflect_call')) {
+        // if Reflect in imports namedImports will be an initialized array with
+        // values, as Reflect can't be constructed and won't be default imported
+        imports.Reflect.namedImports.push('call as Reflect_call');
+      }
+
+      if (imports[mod].namedImports.includes(identifier)) {
+        // if an import of original from module already exists, noop
+        return;
+      }
+
+      imports[mod].namedImports.push(importText);
+      return;
+    default:
+      throw new Error('not yet implemented');
+  }
 }
 
 /**
--- a/lib/imports.js
+++ b/lib/imports.js
@@ -34,7 +34,7 @@ function generateImportText(identifier, type, mod) {
 function updateImports(mod, identifier, type, imports) {
   const importText = generateImportText(identifier, type, mod);
 
-  // TODO(jackbsteinberg): update this for other cases
+  // TODO(jackbsteinberg): update this for other call-requiring cases
   // add original for `call` from `Reflect` upon first method call
   if (type === 'method' && !imports.Reflect) {
     imports.Reflect = {IMPORTS: ['call as Reflect_call']};
--- a/lib/imports.js
+++ b/lib/imports.js
@@ -68,7 +68,7 @@ function addAllImports(sourceFile, imports) {
     Object.entries(imports)) {
     sourceFile.addImportDeclaration({
       defaultImport,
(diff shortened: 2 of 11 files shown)


lib/__testfixtures__/property-access.output.js
```diff
@@ -0,0 +1,5 @@
+import { call as Reflect_call } from "std:global/Reflect";
```
domenic: Great test! The fact that it doesn't change .length is really important, but I totally forgot that was something we had to worry about. I'm glad that this worked out!

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Adds an import for the module to the list
+ *
+ * @param {string} mod - the name of the module to import from
+ * @param {string} identifier - the name of the method / function to get
+ * @param {string} type - the type of original of the identifier
+ * @param {Object} imports - the current list of imports
+ */
+function updateImports(mod, identifier, type, imports) {
```
domenic: Nit: I think putting imports as the first argument would make more sense, since it's what you're updating on. `updateImports(imports, ... stuff to use in updating ...)` reads more naturally than `updateImports(... stuff ..., imports)`.

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+  const importText = generateImportText(identifier, type, mod);
+
+  // TODO(jackbsteinberg): update this for other call-requiring cases
+  // add original for `call` from `Reflect` upon first method call
```
domenic: Is this TODO still accurate?

It seems like the case that's missing is what happens if Reflect already exists. For example an input file like

```js
Reflect.has({}, "prop");
const x = document.createElement('div');
```

or even

```js
Reflect.call(() => {});
const x = document.createElement('div');
```

But the TODO doesn't seem very related to those.

It might be good to add tests for those cases, although getting them right might want to wait until you do static and namespace methods.

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+  if (type === 'method' && !imports.Reflect) {
+    imports.Reflect = {NAMED_IMPORTS: ['call as Reflect_call']};
```
domenic: Use normal camel case; no need for the SHOUTY_CASE (for both NAMED_IMPORTS and DEFAULT).

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  if (mod in imports) {
+    if (type === 'constructor') {
+      imports[mod].DEFAULT = importText;
+      return;
+    }
+
+    if (identifier in imports[mod].NAMED_IMPORTS) {
```
domenic: This will never be true, because NAMED_IMPORTS is an array, so it will only contain numeric keys.

This indicates this code path is not being tested. Write a failing test that exercises what you're trying to do here, then fix the code. (Writing the test first is pretty important here.)

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+      DEFAULT: importText,
+      NAMED_IMPORTS: [],
+    };
+    return;
+  }
+
+  // if no import from module exists, add a new import
+  // of identifier, from module
+  imports[mod] = {NAMED_IMPORTS: [importText]};
+}
```
domenic: I think the structure of this is getting a bit hard to grasp. I would suggest breaking it down like so:

- if mod not in imports
  - create empty structure with no default or named imports
- switch on type:
  - if constructor: do something
  - if method: do something
  - default: throw not yet implemented

lib/imports.js
```diff
@@ -0,0 +1,93 @@
...
+ * Adds all imports from the list to the top of the source file
+ *
+ * @param {SourceFile} sourceFile - the file to add the imports to
+ * @param {Object} imports - the final list of imports
+ */
+function addAllImports(sourceFile, imports) {
+  for (const [MODULE, {DEFAULT, NAMED_IMPORTS}] of Object.entries(imports)) {
+    sourceFile.addImportDeclaration({
+      defaultImport: DEFAULT,
+      namedImports: (NAMED_IMPORTS || !NAMED_IMPORTS.length) ?
```
domenic: It'd be good to structure the code so that NAMED_IMPORTS is always an array. (My above suggestion should help with that.) In that case you can just test the length. (Or maybe you can just join() it anyway and it would work? Worth trying.)

lib/mods/replace-methods.js
```diff
@@ -0,0 +1,47 @@
...
+const {updateImports} = require('../imports.js');
+const {
+  isGlobal,
+  getOriginal,
+} = require('../originals-helper.js');
+
+/**
+ * Builds a call to original `call` from the components of a method call
+ *
+ * @param {*} base - base upon which to call method
```
domenic: What types are these?

lib/__testfixtures__/property-access.output.js
jackbsteinberg: Thanks

lib/imports.js
jackbsteinberg: So I tried rephrasing this TODO but the language is definitely confusing, what I was trying to capture here is that there may be other `type`s that need to use `call` beyond `method`, like for instance `static`. Is that accurate?

I'll go ahead and fix this problem you mentioned as well.

domenic: I guess in that case I'd remove it, since it's not a problem with the current code, just a task for future work.

jackbsteinberg: So if I'm not meant to prepare for the future by writing code future-proofing my code in ways that don't have to do with the current change, and I'm not meant to put comments in reminding myself of the future work I'm meant to do, how do I keep things from slipping through the cracks as I notice them?

domenic: That's what I use the issue tracker for. In general code that gets merged to master should not be used as an issue tracker or to-do list; it should be something folks can read and understand in isolation as "what works now", not a work in progress.

lib/imports.js
jackbsteinberg: I will go ahead and fix this, but it actually doesn't seem to matter, the tests won't fail because somewhere in the process of emitting the project to a file in memory the duplicate imports get cleaned up. I added a line to another test that would test this if the emitting didn't just fix it for me

lib/imports.js
jackbsteinberg: I'll go ahead and fix it so it's always an array, and after checking a bit it looks like I can join anyway, passing an empty string to namedImports doesn't add any

jackbsteinberg: Hmm it looks like I don't even need join, passing in an array of strings seems to work just fine

lib/imports.js
```diff
@@ -0,0 +1,89 @@
...
+    case 'constructor':
+      imports[mod].defaultImport = importText;
+      return;
+    case 'method':
+      if (!imports.Reflect) {
+        imports.Reflect = {namedImports: ['call as Reflect_call']};
+      } else if (!imports.Reflect.namedImports
+          .includes('call as Reflect_call')) {
+        // if Reflect in imports namedImports will be an initialized array with
+        // values, as Reflect can't be constructed and won't be default imported
```
domenic: Well, it'll always be an initialized array, because of line 39.

lib/imports.js
```diff
@@ -0,0 +1,89 @@
...
+      namedImports: [],
+    };
+  }
+
+  switch (type) {
+    case 'constructor':
+      imports[mod].defaultImport = importText;
+      return;
+    case 'method':
+      if (!imports.Reflect) {
```
domenic: It might be simpler if you made namedImports a Set. Then you could just add() to it and not worry about if it already exists (because adding something that already exists to a Set is a no-op). This might require converting it to an array (with `[...set]`) before passing it to ts-morph, but I think it would clean up some of the code here a decent bit.

lib/imports.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Adds all imports from the list to the top of the source file
+ *
+ * @param {SourceFile} sourceFile - the file to add the imports to
+ * @param {Object} imports - the final list of imports
+ */
+function addAllImports(sourceFile, imports) {
+  for (const [MODULE, {defaultImport, namedImports}] of
```
domenic: Should not be uppercase

lib/imports.js
```diff
@@ -0,0 +1,89 @@
...
+    Object.entries(imports)) {
+    sourceFile.addImportDeclaration({
+      defaultImport: defaultImport,
```
domenic: Can just be `defaultImport`

lib/imports.js
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * @param {SourceFile} sourceFile - the file to add the imports to
+ * @param {Object} imports - the final list of imports
+ */
+function addAllImports(sourceFile, imports) {
+  for (const [moduleName, {defaultImport, namedImports}] of
+    Object.entries(imports)) {
+    sourceFile.addImportDeclaration({
+      defaultImport,
+      namedImports: [...namedImports],
```
domenic: I'm curious, does it work if you just pass the set directly? If so that'd be nice.

jackbsteinberg: Well I'll be, it totally does - that's nice

jackbsteinberg: Wait never mind I was on the wrong branch when I tested, this doesn't work :/

lib/imports.js
fergald: In general try to avoid TODO(person) because then you're using the code as an issue tracker. It does sometimes make sense to have TODO(ISSUE-URL), e.g. when you want to explain why something is wrong or missing in the code.

PR #22, opened 2019-08-07
Build support for statics into methods mod & switch to using `apply`

jackbsteinberg:
After a bit of digging I figured out that static methods aren't such a different case from normal methods, so they can be easily combined into the methods mod, which is what I did here. Additionally I changed from using `Replace_call`, a method that doesn't exist, to `Replace_apply`, which does.

### DIFF
--- a/lib/imports.js
+++ b/lib/imports.js
@@ -19,7 +19,7 @@ function generateImportText(identifier, type, mod) {
     case 'method':
       return `${identifier} as ${original}`;
     case 'static':
-      return `${identifier}`;
+      return `${identifier}_static`;
     default:
       throw new Error('not yet implemented');
   }
@@ -48,19 +48,20 @@ function updateImports(imports, mod, identifier, type) {
       return;
     case 'method':
       if (!imports.Reflect) {
-        imports.Reflect = {namedImports: new Set(['call as Reflect_call'])};
+        imports.Reflect = {namedImports: new Set(['apply as Reflect_apply'])};
       }
 
-      imports.Reflect.namedImports.add('call as Reflect_call');
+      imports.Reflect.namedImports.add('apply as Reflect_apply');
       imports[mod].namedImports.add(importText);
       return;
     case 'static':
       if (!imports.Reflect) {
-        imports.Reflect = {namedImports: new Set(['call as Reflect_call'])};
+        imports.Reflect = {namedImports: new Set(['apply as Reflect_apply'])};
       }
 
-      imports.Reflect.namedImports.add('call as Reflect_call');
+      imports.Reflect.namedImports.add('apply as Reflect_apply');
       imports[mod].namedImports.add(importText);
+      imports[mod].defaultImport = mod;
       return;
     default:
       throw new Error('not yet implemented');
@@ -76,7 +77,6 @@ function updateImports(imports, mod, identifier, type) {
 function addAllImports(sourceFile, imports) {
   for (const [moduleName, {defaultImport, namedImports}] of
     Object.entries(imports)) {
-    console.log(moduleName);
     sourceFile.addImportDeclaration({
       defaultImport,
       namedImports: [...namedImports],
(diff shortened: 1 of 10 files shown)


lib/imports.js
```diff
@@ -46,10 +48,10 @@ function updateImports(imports, mod, identifier, type) {
       return;
     case 'method':
       if (!imports.Reflect) {
-        imports.Reflect = {namedImports: new Set(['call as Reflect_call'])};
+        imports.Reflect = {namedImports: new Set(['apply as Reflect_apply'])};
       }
-
-      imports.Reflect.namedImports.add('call as Reflect_call');
+      imports.Reflect.namedImports.add('apply as Reflect_apply');
+    case 'static':
```
domenic: Don't use switch fall-through in this way. Put the `break` before the `case`, and then duplicate the `imports[mod].namedImports.add(importText);` line. (In this case I think clarity outweighs avoiding duplication.)

Also it would be good to switch from `return` to `break` in general.

lib/mods/replace-methods.js
```diff
@@ -30,17 +38,25 @@ function replaceGlobalMethodCalls(node, imports) {
   if (nodeTypes.isPropertyAccessExpression(node) &&
       nodeTypes.isCallExpression(node.getParent()) &&
       isGlobal(node.getExpression())) {
-    const callerType = node.getExpression().getType().getSymbol().getName();
+    let callerType = node.getExpression().getType().getSymbol().getName();
     const method = node.getSymbol().getEscapedName();
-    const original = getOriginal(method, 'method', callerType);
+    let type = 'method';
+
+    if (node.getExpression().getType().getConstructSignatures()[0]) {
```
domenic: Let's try to save variables in expressions a bit more. I think at least we should have `const baseTypeConstructSignature = node.getExpression().getType().getConstructSignatures()[0]`, but maybe we should also have `base` and `baseType`.

lib/imports.js
jackbsteinberg: Yeah I had this trick in when it helped prevent code duplication but it has since outstayed its welcome, fixing now

lib/mods/replace-methods.js
jackbsteinberg: I added a baseType, let me know if you think we need a base

lib/mods/replace-methods.js
```diff
@@ -30,17 +38,26 @@ function replaceGlobalMethodCalls(node, imports) {
   if (nodeTypes.isPropertyAccessExpression(node) &&
       nodeTypes.isCallExpression(node.getParent()) &&
       isGlobal(node.getExpression())) {
-    const callerType = node.getExpression().getType().getSymbol().getName();
+    const baseType = node.getExpression().getType();
+    let callerType = baseType.getSymbol().getName();
```
domenic: I think callerType should be called baseTypeName.

domenic: You could also move this down to after the `if`, and change the `callerType = ` to be `baseType =` and then lose the `.getSymbol().getName()`.

lib/mods/replace-methods.js
```diff
@@ -38,26 +38,27 @@ function replaceGlobalMethodCalls(node, imports) {
   if (nodeTypes.isPropertyAccessExpression(node) &&
       nodeTypes.isCallExpression(node.getParent()) &&
       isGlobal(node.getExpression())) {
-    const baseType = node.getExpression().getType();
-    let callerType = baseType.getSymbol().getName();
+    let baseType = node.getExpression().getType();
```
domenic: This still isn't really what I'm thinking. Now baseType is used for two different purposes: sometimes it's a type, and sometimes it's a name.

Here is my suggestion:

```js
let baseType = node.getExpression().getType();
let baseConstructSignature = baseType.getConstructSignatures()[0];

if (baseConstructSignature) {
  baseType = baseConstructSignature.getReturnType();
}

const baseTypeName = baseType.getSymbol().getEscapedName();
```

jackbsteinberg: Ok got it

jackbsteinberg: Hopefully this resolves it

Issue #26, opened 2019-08-08
Getters/setters


domenic:
Getters/setters are similar to methods; the main difference is that they're a type of property access expression that's not a call expression. Also, being sure to distinguish getter vs. setter cases.

Inputs:

```js
const set = new Set();
set.size;

const url = new URL();
url.href;
url.pathname = "foo";
```

Outputs, ignoring the constructor transforms:

```js
import { apply as Reflect_apply } from "std:global/Reflect";
import { size_get as Set_size_get } from "std:global/Set";
import { href_get as URL_href_get, pathname_set as URL_pathname_set } from "std:global/URL";

const set = new Set();
Reflect_apply(Set_size_get, set); // it's OK to pass [] as last arg, just redundant

const url = new URL();
Reflect_apply(URL_href_get, url);
Reflect_apply(URL_pathname_set, url, ["foo"]);
```

Note: the above examples dodge #25, we'll also want getter/setter tests for those.

domenic:
Important note: this should only apply to accessor properties, not data properties. Notably, `array.length` should not get rewritten at all; it should stay `array.length`.

I'm not sure if there's a more generic detection story. We may end up just special-casing `array.length` as its one of the only data properties in the standard library.

domenic:
I've concluded TypeScript does not the information to tell whether something's an accessor or data property. We should hard-code in array and string length for now, and have a follow-up to handle other cases.

The long-term solution is still probably a hard-coded list. Only the JS spec uses data properties, and it's pretty small relative to the entirety of the web. It's also not in a machine-scrapable format. So we'll just have to have someone (probably me) go through the JS spec and find all the data properties. I would guess there are probably like 5-10 max (with the 2 `length`s being by far the most common).

domenic:
Created #33 to track the above-discussed detection of data properties, so that this issue can focus on the core functionality of accessor replacement (with only the two length properties hard-coded as data properties for now).

jackbsteinberg:
Generic getter / setter case resolved in #28, closing this issue

PR #28, opened 2019-08-08
Add mod for replacing gets/sets with originals

jackbsteinberg:


.gitignore
```diff
@@ -1,3 +1,4 @@
 /node_modules
+.vscode
```
domenic: It's best to add this to your personal global .gitignore. http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

lib/__testfixtures__/get-set.input.js
```diff
@@ -0,0 +1,3 @@
+const arr = [1, 2, 3];
+arr.length = 5;
+const b = arr.length;
```
domenic: So, `arr.length` should specifically not get rewritten. Keep this test, but the expectation should change. Also, add a similar test for string.length (with no setter).

To make the test pass (i.e. not modify those .lengths), add a function called `isDataProperty` which takes some useful info (probably type and property name?). For now, hard-code it to return true for string and array `length` properties. Add a TODO in it to use a more robust mechanism. Then, if `isDataProperty()` returns true, do not perform the transformation.

Also, please include both a JS and web example. #26 has some. (Note: there are no good examples for JS setters, only web setters. That's fine.)

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
...
+} = require('../originals-helper.js');
+const {formatCall} = require('./helpers.js');
+
+/**
+ * Replaces accessor calls in the current node with the Get-Originals version
+ *
+ * @param {Node} node - a ts-morph AST Node
+ * @param {Object} imports - the current list of imports
+ */
+function replaceGlobalGetSetCalls(node, imports) {
```
domenic: Nit: naming is all over the place. In the comments it's accessor calls. In the function name it's global get set calls. In the filename and the export of index.js it's get set. Pick one :). I like get set or maybe "gets and sets".

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
...
+      isGlobal(node.getExpression())) {
+    const method = node.getSymbol().getEscapedName();
+    const baseTypeName = node.getExpression().getType().getSymbol().getName();
+    const callee = node.getExpression().getSymbol().getName();
+    let type;
+    let args;
+
+    // Check if the expression is receiving assignment (set)
+    if (nodeTypes.isBinaryExpression(node.getParent()) &&
+        node.getParent().getLeft() === node) {
```
domenic: I think this will give a false positive on cases like `url.pathname && 5;` (it will think it's an assignment). Add a test for that, and if it fails, find a stricter check.

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
...
+      type = 'set';
+      args = node.getParent().getRight().getText();
+    } else {
+      type = 'get';
+    }
+
+    const original = getOriginal(method, type, baseTypeName);
+
+    let base;
```
domenic: I don't think "base" is quite right here. You're potentially replacing a whole node, not just the base portion. `nodeToReplace` seems more accurate.

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const nodeTypes = require('ts-morph').TypeGuards;
+const {updateImports} = require('../imports.js');
+const {
+  isGlobal,
+  getOriginal,
+} = require('../originals-helper.js');
+const {formatCall} = require('./helpers.js');
```
domenic: formatCall no longer seems like an accurate name. formatReplacement?

lib/mods/helpers.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
```
domenic: What's the difference between originals-helper.js and helpers.js?

lib/rewriter.js
```diff
@@ -3,10 +3,13 @@
...
-const cleanConsole = require('./mods/clean-console.js');
-const replaceConstructors = require('./mods/replace-constructors.js');
-const replaceMethods = require('./mods/replace-methods.js');
 const {addAllImports} = require('./imports.js');
+const {
+  cleanConsole,
+  replaceConstructors,
+  replaceMethods,
+  replaceGetSet,
+} = require('./mods');
```
domenic: The index.js file seems like unnecessary redirection here. It just added code (the require statements still exist, over there).

I'd suggest just continuing with the current architecture. We could consider something where index.js contains an array of functions, and then this file loops over them and applies them all; that would potentially be a worthwhile abstraction. But let's do that as a separate PR, if we want to.

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
...
+ * Replaces accessor calls in the current node with the Get-Originals version
+ *
+ * @param {Node} node - a ts-morph AST Node
+ * @param {Object} imports - the current list of imports
+ */
+function replaceGlobalGetSetCalls(node, imports) {
+  if (nodeTypes.isPropertyAccessExpression(node) &&
+      !nodeTypes.isCallExpression(node.getParent()) &&
+      isGlobal(node.getExpression())) {
+    const method = node.getSymbol().getEscapedName();
```
domenic: method isn't quite right. I'd go with accessorName or propertyName.

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,50 @@
...
+    const baseTypeName = node.getExpression().getType().getSymbol().getName();
+    const callee = node.getExpression().getSymbol().getName();
```
domenic: I think this is the base. Callee is weird when nobody is being called.

lib/mods/replace-get-set.js
jackbsteinberg: I'm just gonna go ahead and use `getset`, I stumbled on the name accessor calls late in the game so I added a few of those, but it's somewhat less explicit

lib/mods/helpers.js
jackbsteinberg: Very little it turns out, I could easily put formatReplacement into `originals-helper.js` and I think I'm gonna go ahead and do that

lib/__testfixtures__/get-set.input.js
jackbsteinberg: For arr.length not being rewritten, in the case of something like:
```
arr.length = 5;
```
Should I keep it as is? Or should I change it to:
```
Reflect_apply(Array_length_get, arr) = 5;
```

I currently have it as the former, but could easily switch to the latter

domenic: You should keep it as is. You cannot assign to the result of a function.

jackbsteinberg: Yeah that’s what I expected, just wanted to be sure

lib/__testfixtures__/get-set.output.js
```diff
@@ -0,0 +1,8 @@
+import URL, { pathname_get as URL_pathname_get } from "std:global/URL";
+import { length_get as Array_length_get } from "std:global/Array";
+import { apply as Reflect_apply } from "std:global/Reflect";
+const arr = [1, 2, 3];
+arr.length = 5;
+const b = Reflect_apply(Array_length_get, arr);
```
domenic: This should not be rewritten.

lib/__testfixtures__/property-access.output.js
```diff
@@ -1,5 +1,5 @@
-import { toString as Array_toString } from "std:global/Array";
+import { toString as Array_toString, length_get as Array_length_get } from "std:global/Array";
 import { apply as Reflect_apply } from "std:global/Reflect";
 const a = [1, 2, 3];
-const l = a.length;
+const l = Reflect_apply(Array_length_get, a);
```
domenic: This should not be rewritten.

lib/originals-helper.js
```diff
@@ -29,7 +29,7 @@ function nodeInIDLSpecs(node) {
  * @return {boolean} whether the namespace is global
  */
 function isGlobal(node) {
-  return nodeInTSDefs(node) || nodeInIDLSpecs(node);
+  return nodeInIDLSpecs(node) || nodeInTSDefs(node);
```
domenic: Why change this?

lib/__testfixtures__/get-set.output.js
jackbsteinberg: So for data properties I shouldn't be rewriting getters or setters? I thought it was only setters previously

domenic: No, it's both. For data properties there are no originals.

lib/originals-helper.js
jackbsteinberg: I wanted to guard against the possibility that a Web type exists in the IDL but the nodes for the TypeScript check don't exist, which would throw an error. I could try / catch it also, but this seemed easier

domenic: The function shouldn't throw an error if the node is not in the TypeScript defs. It should instead return false. (And this should be done without try/catch, e.g. by detecting something is undefined instead of accessing a property of undefined.)

lib/__testfixtures__/get-set.input.js
```diff
@@ -0,0 +1,5 @@
...
+const u = new URL();
+const c = u.pathname && 5;
```
domenic: Can you test assigning to url.pathname? There are no tests for sets being rewritten anymore.

domenic: Can you test a JS get, like the one listed in #26? (`Set`'s `size`)

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,74 @@
...
+} = require('../originals-helper.js');
+
+const dataProperties = new Map([
+  ['Array', ['length']],
+  ['String', ['length']],
+]);
+
+/**
+ * @param {*} type - the underlying type of the object
+ * @param {*} propertyName - the name of the property
```
domenic: What types are these arguments and return types?

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    const base = node.getExpression().getSymbol().getName();
+    let type;
+    let args;
+    const parent = node.getParent();
+
+    // Check if the expression is receiving assignment (set)
+    if (nodeTypes.isBinaryExpression(parent) &&
+    parent.getOperatorToken().getText() === '=' &&
```
domenic:
```suggestion
        parent.getOperatorToken().getText() === '=' &&
```

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,74 @@
...
+        node.getParent().getLeft() === node) {
+      type = 'set';
+      args = parent.getRight().getText();
+    } else {
+      type = 'get';
+    }
+
+    const original = getOriginal(propertyName, type, baseTypeName);
+
+    let nodeToReplace;
```
domenic: This seems simpler if you incorporate it into the above if.

lib/originals-helper.js
domenic: Let's put this back to the original version? If we think the other order is better, that should be a separate PR.

lib/rewriter.js
```diff
@@ -3,10 +3,11 @@
 const assert = require('assert');
 const ts = require('typescript');
 const tsMorph = require('ts-morph');
+const {addAllImports} = require('./imports.js');
```
domenic: Nit: if you move this to the bottom then the diff is simpler. (I.e. then this PR doesn't move it around.)

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Replaces getset calls in the current node with the Get-Originals version
+ *
+ * @param {Node} node - a ts-morph AST Node
+ * @param {Object} imports - the current list of imports
+ */
+function replaceGlobalGetSetCalls(node, imports) {
+  if (nodeTypes.isPropertyAccessExpression(node) &&
+      !nodeTypes.isCallExpression(node.getParent()) &&
```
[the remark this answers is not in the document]
jackbsteinberg: This helps us see that it's not a method call, since they need to be treated differently from get-sets

```
// get/set
url.pathname;

// method call
url.toString();
```

lib/mods/replace-get-set.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+ * Replaces getset calls in the current node with the Get-Originals version
+ *
+ * @param {Node} node - a ts-morph AST Node
+ * @param {Object} imports - the current list of imports
+ */
+function replaceGlobalGetSetCalls(node, imports) {
+  if (nodeTypes.isPropertyAccessExpression(node) &&
```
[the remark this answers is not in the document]
jackbsteinberg: Out of date as of #43

lib/mods/replace-get-set.js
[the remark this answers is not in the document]
fergald: I'm still confused by this, the example provided is not against a global but I think maybe I understand now, you are trying to prevent `window.alert()` from becoming `Reflect_apply(Window_alert_get, window)()` ?

Is that correct? I don't see something in the get_set test that looks like the above, so maybe I've misunderstood.

Anyway, whatever the explanation is, it should be inline in the code. E.g.

```js
  // Must be a property access on a global, e.g. window.location
  if (!(nodeTypes.isPropertyAccessExpression(node) &&
      isGlobal(node.getExpression()))) {
    return;
  }

  // Must not be a call e.g. window.alert().
  if (nodeTypes.isCallExpression(node.getParent())) {
    return;
  }
```

Issue #24, opened 2019-08-08
Namespace functions


domenic:
We currently correctly transform static functions, i.e. functions that exist on a class, like `Promise.all()` or `Array.isArray()`. There's a related-but-different class of functions, which are functions that exist on a namespace. Examples include `Math.max()`, `Reflect.apply()`, and `CSS.escape()`.

Note that `Reflect.apply()` may be a tricky case because we're already using it in some ways.

Input:

```js
Math.max();
Reflect.apply();
CSS.escape();
```

output:

```js
import { max as Math_max } from "std:global/Math";
import { apply as Reflect_apply } from "std:global/Reflect";
import { escape as CSS_escape } from "std:global/CSS";

Math_max();
Reflect_apply();
CSS_escape();
```

The unknown here is how TypeScript/ts-morph views the bases, and how it distinguishes them from things like `Array` or `Promise` or `URL`.

PR #34, opened 2019-08-09
Add mod for replacing namespace methods (Issue #24)

jackbsteinberg:


lib/mods/replace-methods.js
```diff
@@ -8,6 +8,24 @@ const {
...
+/**
+ *
+ * @param {Node} node - the node to examine
+ * @param {string} baseTypeName - the base type of the node
+ *
+ * @return {boolean} whether the method is a namespace method
+ */
+function isNamespaceMethod(node, baseTypeName) {
+  if (node.getExpression().getSymbol().getDeclarations()[0]
+      && node.getExpression().getSymbol().getDeclarations()[0]
```
domenic: Go for [DRY code](https://www.softwareyoga.com/is-your-code-dry-or-wet/). Save this in a variable instead of repeating it three times.

lib/mods/replace-methods.js
```diff
@@ -8,6 +8,24 @@ const {
...
+          .getInitializer) {
```
domenic: This will always be true, because the function always exists. I think you want to test `.getInitializer()`.

However, the fact that this bug is passing the tests, means that either this second check is not important, or the tests are not thorough enough. I suggest deleting this second check for now, unless you have some test that will fail without it.

lib/mods/replace-methods.js
```diff
@@ -8,6 +8,24 @@ const {
...
+    const callee = node.getExpression().getSymbol().getDeclarations()[0]
+        .getInitializer().getText();
+    return callee === baseTypeName;
+  }
+  return node.getExpression().getSymbol().getName() === baseTypeName;
```
domenic: Similarly, node.getExpression().getSymbol() should be saved to a variable.

lib/mods/replace-methods.js
jackbsteinberg: So without the second check I'm getting the error:

```
TypeError: firstDeclaration.getInitializer is not a function
```

on the line

```
const callee = firstDeclaration.getInitializer().getText();
```

I don't think the function always exists if firstDeclaration does, which is why I have that second check

domenic: Fascinating. What is firstDeclaration in that case?

jackbsteinberg: Seems to be an `InterfaceDeclaration` for `Math`

Issue #30, opened 2019-08-09
Remove jscodeshift dependency


domenic:
Since we're now using ts-morph, we shouldn't bring in the large jscodeshift dependency. Currently we are only using it for being a command-line wrapper that feeds `rewriter.js` the source text, plus we are using its testing framework.

For the command-line wrapper, I think we should replace it with our own custom CLI. This is not too hard to write. The usage should be

```bash
$ node rewriter.js input.mjs output.mjs
```

(I don't think we need to support multiple inputs/outputs; just run the script multiple times.) An example of how this could work is in https://github.com/domenic/svg2png/blob/master/bin/svg2png-cli.js, although we don't need the complicated args parsing using yargs.

For the tests, I think this might be a good match for my [baseline-tester](https://www.npmjs.com/package/baseline-tester) project, or we could use Jest snapshot testing.

domenic:
Discussed in person: let's keep using Jest, instead of moving to baseline tester. We may or may not want to use snapshot testing; I like the current straightforward input.js/output.js structure instead of the single snapshot file.

domenic:
Also, this task is getting bumped up in priority, because currently we broke the command line invocation.

PR #36, opened 2019-08-12
Add mod for replacing Window methods and explicit get-sets

### LINKED ISSUE #21: Window methods
Window methods like `setTimeout()` are going to cause difficulty because they don't follow the same pattern as other method calls and won't get detected and modified. 

Additionally, Window isn't in the `globals-set.js` file to begin with, so `isGlobal(Window)` would return false anyway.

This is looking like it's going to have to be its own separate case to be handled.

jackbsteinberg:
Fixes #21


lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @param {*} node - the node to examine
+ *
+ * @return {boolean} whether the node represents a window function
+ */
+function isWindowProperty(node) {
+  const symbol = node.getExpression().getSymbol();
+  const firstDeclaration = (symbol) ? symbol.getDeclarations()[0] : undefined;
+
+  // TODO: may require a deeper check
```
jackbsteinberg: This should be taken out before merge, but I wanted to highlight the fact that this could require looking deeper - I'm not sure what test would show this though

lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,57 @@
...
+
+const nodeTypes = require('ts-morph').TypeGuards;
+const {updateImports} = require('../imports.js');
+const {
+  formatReplacement,
+  getOriginal,
+} = require('../originals-helper.js');
+
+/**
+ * @param {*} node - the node to examine
```
domenic: What type is this? It'd be good to do a first pass before sending things for review to catch this sort of thing.

lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @return {boolean} whether the node represents a window function
+ */
+function isWindowProperty(node) {
+  const symbol = node.getExpression().getSymbol();
+  const firstDeclaration = (symbol) ? symbol.getDeclarations()[0] : undefined;
```
domenic:
```suggestion
  const firstDeclaration = symbol && symbol.getDeclarations()[0];
```

lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,57 @@
...
+
+  // TODO: may require a deeper check
+  // e.g. setTimeout web returns number, Node returns Timer
+  if (firstDeclaration) {
+    return window.hasOwnProperty(firstDeclaration.getName());
```
domenic: What is `window` here? `window` shouldn't exist in Node.js... Maybe the testing environment is introducing it!? That'd be bad because it'd fail outside of our tests.

lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {*} node - the node to examine
+ *
+ * @return {boolean} whether the node represents a window function
+ */
+function isWindowProperty(node) {
+  const symbol = node.getExpression().getSymbol();
+  const firstDeclaration = (symbol) ? symbol.getDeclarations()[0] : undefined;
+
+  // TODO: may require a deeper check
+  // e.g. setTimeout web returns number, Node returns Timer
```
domenic: This doesn't really clarify it for me. We're never operating in Node, only the web. What is a case where firstDeclaration is undefined vs. not?

lib/mods/replace-window-functions.js
jackbsteinberg: I wasn't really sure what this was or how it got here either but it seemed to work.

I could look into alternatives for this if we think it's gonna fail outside of the tests (which I thought were running in a node environment but maybe that's a jest config problem)

domenic: Yeah, I think that's a Jest config problem. You can try by running the command from the readme (`npm run rewrite ...`).

lib/mods/replace-window-functions.js
jackbsteinberg: I think I just got confused here, it seemed too easy (because it was, I can't just use `window` the way I currently am)

lib/mods/replace-window-functions.js
jackbsteinberg: So I swapped this here, but I have yet to run the tests in a node environment because running jscodeshift is broken - will be fixed as I resolve #30 in an upcoming PR

lib/mods/replace-window-functions.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+function replaceWindowFunctions(node, imports) {
+  if (nodeTypes.isCallExpression(node) && isGlobal(node.getExpression())) {
+    const symbol = node.getExpression().getSymbol();
+    const firstDeclaration = symbol.getDeclarations()[0];
+    const type = 'window';
+    const mod = 'Window';
+    let name = symbol.getName();
+
+    if (firstDeclaration) {
```
domenic: When is this necessary?

jackbsteinberg: According to jest's `--coverage` flag, never

lib/mods/replace-methods.js
```diff
@@ -54,34 +54,47 @@ function isNamespaceMethod(node, baseTypeName) {
...
-      callee = baseTypeName;
-    }
+      if (baseConstructSignature) {
+        baseType = baseConstructSignature.getReturnType();
+        type = 'static';
+      }
+
+      mod = baseType.getSymbol().getEscapedName();
+      args = propertyAccess.getParent().getArguments()
+          .map((elt) => elt.getText()).join(',');
```
fergald: Seems like

```js
function argumentString(functionCall) {
  return functionCall.getArguments().map((elt) => elt.getText()).join(',');
}
```

would dedupe some code here and below.

lib/mods/replace-methods.js
```diff
@@ -54,34 +54,47 @@ function isNamespaceMethod(node, baseTypeName) {
...
 function replaceGlobalMethodCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      nodeTypes.isCallExpression(node.getParent()) &&
+  if (nodeTypes.isCallExpression(node) &&
       isGlobal(node.getExpression())) {
-    const method = node.getSymbol().getEscapedName();
-    let type = 'method';
-    let baseType = node.getExpression().getType();
-    const baseConstructSignature = baseType.getConstructSignatures()[0];
+    let mod; let args; let type; let callee;
```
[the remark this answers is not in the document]
jackbsteinberg: That's what I had initially but ES-Lint didn't seem to be happy with it, it told me to fix it by splitting it into individual lets. I can change the rule here if we have a strong preference, but for now the linter doesn't allow it.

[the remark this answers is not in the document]
domenic: In particular the Google style guide prohibits Fergal's suggestion: https://google.github.io/styleguide/jsguide.html#features-local-variable-declarations

lib/mods/replace-methods.js
```diff
@@ -47,41 +47,61 @@ function isNamespaceMethod(node, baseTypeName) {
...
-    const original = getOriginal(method, type, baseTypeName);
+      if (type !== 'static' && isNamespaceMethod(propertyAccess, mod)) {
+        type = 'namespace';
+        callee = mod;
+      }
+    } else { // window functions
+      type = 'window';
+      mod = 'Window';
+      name = node.getExpression().getSymbol()
+          .getDeclarations()[0].getName();
```
domenic: Can you just do `node.getExpression().getSymbol()..getName()` like the above?

lib/mods/replace-methods.js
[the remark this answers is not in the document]
fergald: Oh, C++ style is fine with it and I see it used a bit (without any assignment of initial values) but yeah, JS-style doesn't like it. JS-style also prohibits multiple statements on 1 line. So the style-compliant version of this is

let mod;
let args;
let type;
let callee;

Issue #38, opened 2019-08-13
Finish, remove, or replace IDL-based global detection


domenic:
For constructors, we use data from the reffy-reports crawl.json as a backstop, to determine if something is a global constructor. That is, first we check the TypeScript definitions, and then we check the crawl.json derived [`globals-set.js`](https://github.com/jackbsteinberg/get-originals-rewriter/blob/master/lib/globals-set.js).

The idea here was that the TypeScript definitions are updated less frequently than the universe of web specifications, so this would catch some cases that the TypeScript definitions would not.

We have not done similar work for other transforms (methods, getters/setters, namespace methods, ...).

We should either:

- Extend this infrastructure to the other transforms. E.g., create a list of all methods on all web classes, and use that in the methods transform.
- Remove this infrastructure and replace it with something simpler. For example, a "additional-originals.js" file, either checked in to this repository or fed as input to the CLI tool, which can contain any specific originals that we need to rewrite but have not made it in to the TypeScript definitions yet.
- Remove this infrastructure and replace it with something TypeScript based, e.g. allow additional .d.ts files.

I tend toward remove/replace. The last option seems most congruent with our existing infrastructure, but a bit more work. The second option is probably simplest.

I think it would be fine to do this in two parts, first remove, then replace. We could even wait until we have a concrete case of something not in the TypeScript definitions before working on the replacement.

fergald:
I wonder if

https://cs.chromium.org/chromium/src/third_party/blink/web_tests/webexposed/global-interface-listing-expected.txt

would be helpful here. Obviously not something we can include in this repo but maybe we can teach this to load a file in that format and then when used in blink, we can pass this file.

jackbsteinberg:
Just pushed #47 to remove IDL checking, the issue still stands at the moment however with regards to newer globals that TypeScript will take longer to get around to, so there need will be a replacement for the IDL checking

PR #39, opened 2019-08-13
Update lib/globals-set.js due to crawl.json fix

jackbsteinberg:
`crawl.json` errors fixed in [tidoust/reffy-reports/#36](https://github.com/tidoust/reffy-reports/issues/36), requiring an update to the tracked version of globals-set.js


domenic:
I'm not sure we should stop tracking it, given that it's historically been unstable; it would be good to see changes evolve over time. Updating it seems reasonable but if we're still using it, keeping it tracked for a while makes sense to me.

Alternately, we could just remove this part of the infrastructure entirely, as mentioned in #38.

jackbsteinberg:
I'm ok with continuing to track it, in which case this would just be an update to it. I'm gonna go ahead and change to that now.

PR #43, opened 2019-08-19
Refactor Get-Set Replacement mod for clarity, structure, and pure downstream node replacement

jackbsteinberg:
Changes the `replaceWithText` target from the node's parent to the node or its children to keep the traversal simplified.

### DIFF
--- a/lib/mods/replace-get-set.js
+++ b/lib/mods/replace-get-set.js
@@ -50,11 +50,10 @@ function isGlobalSet(node) {
 /**
  * Replaces getset calls in the current node with the Get-Originals version
  *
- * @param {Node} node - a ts-morph AST Node
+ * @param {Node} node - a ts-morph AST Nodec
  * @param {Object} imports - the current list of imports
  */
 function replaceGlobalGetSetCalls(node, imports) {
-  const a = node.getText();
   if (isGlobalGet(node)) {
     const propertyName = node.getSymbol().getEscapedName();
     const baseTypeName = node.getExpression().getType().getSymbol().getName();
--- a/lib/mods/replace-get-set.js
+++ b/lib/mods/replace-get-set.js
@@ -25,15 +25,37 @@ function isDataProperty(type, propertyName) {
 }
 
 /**
+ * @param {Node} node - the current ts-morph AST Node
+ *
+ * @return {boolean} whether the node is a global get
+ */
+function isGlobalGet(node) {
+  return nodeTypes.isPropertyAccessExpression(node) &&
+    !nodeTypes.isCallExpression(node.getParent()) &&
+    isGlobal(node.getExpression());
+}
+
+/**
+ * @param {Node} node - the current ts-morph AST Node
+ *
+ * @return {boolean} whether the node is a global set
+ */
+function isGlobalSet(node) {
+  return nodeTypes.isBinaryExpression(node) &&
+    node.getOperatorToken().getText() === '=' &&
+    nodeTypes.isPropertyAccessExpression(node.getLeft()) &&
+    isGlobal(node.getLeft().getExpression());
+}
+
+/**
  * Replaces getset calls in the current node with the Get-Originals version
  *
  * @param {Node} node - a ts-morph AST Node
  * @param {Object} imports - the current list of imports
  */
 function replaceGlobalGetSetCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      !nodeTypes.isCallExpression(node.getParent()) &&
-      isGlobal(node.getExpression())) {
+  const a = node.getText();
+  if (isGlobalGet(node)) {
     const propertyName = node.getSymbol().getEscapedName();
     const baseTypeName = node.getExpression().getType().getSymbol().getName();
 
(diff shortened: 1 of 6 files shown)


lib/mods/replace-get-set.js
```diff
@@ -24,16 +24,37 @@ function isDataProperty(type, propertyName) {
...
  * Replaces getset calls in the current node with the Get-Originals version
  *
  * @param {Node} node - a ts-morph AST Node
  * @param {Object} imports - the current list of imports
  */
 function replaceGlobalGetSetCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      !nodeTypes.isCallExpression(node.getParent()) &&
-      isGlobal(node.getExpression())) {
+  if (isGlobalGet(node)) {
```
domenic: After this refactoring, there is a lot of duplicated code between the two branches. Is it possible to reduce that? We could do it in a separate PR if you'd prefer.

lib/mods/replace-methods.js
```diff
@@ -30,7 +30,8 @@ function isNamespaceMethod(node, baseTypeName) {
    *
    * First declared as CSS, with a baseTypeName of CSS (Namespace ✔)
    */
-  if (firstDeclaration && firstDeclaration.getInitializer) {
+  if (firstDeclaration && firstDeclaration.getInitializer
+        && firstDeclaration.getInitializer()) {
```
domenic: Why was this change introduced in this PR?

lib/mods/replace-methods.js
```diff
@@ -54,12 +55,14 @@ function isNamespaceMethod(node, baseTypeName) {
...
  */
 function replaceGlobalMethodCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      nodeTypes.isCallExpression(node.getParent()) &&
-      isGlobal(node.getExpression())) {
-    const method = node.getSymbol().getEscapedName();
+  if (node.getExpression &&
+      nodeTypes.isPropertyAccessExpression(node.getExpression()) &&
+      nodeTypes.isCallExpression(node) &&
+      isGlobal(node.getExpression().getExpression())) {
```
domenic: I think if you move `nodeTypes.isCallExpression()` first, then you will not need to test `node.getExpression`.

Also, it would be good to factor this out into `isGlobalMethodCall` like you did for get/set.

lib/originals-helper.js
```diff
@@ -22,8 +22,12 @@ function nodeInTSDefs(node) {
  * @return {boolean} whether the type of the node matches IDL in a web spec
  */
 function nodeInIDLSpecs(node) {
-  return globalsSet.has(
-      node.getType().getSymbol().getName());
+  const symbol = node.getType().getSymbol();
```
domenic: Why was this change introduced in this PR?

lib/mods/replace-methods.js
jackbsteinberg: Will move to refactor PR

lib/originals-helper.js
jackbsteinberg: Will move to refactor PR

jackbsteinberg:
After rebasing and cleaning out things that belong in a general refactor, I realize this PR really just serves as a general refactor to the `replace-get-set.js` file. I think instead of reparenting in this PR and creating another PR to do the refactor I did here, I can pivot this with the purpose of refactoring `replace-get-set.js` - thoughts?

domenic:
That sounds reasonable. However, be sure to distinguish between "refactor" (~ moving code around) and "change" (some of the things I comment on added new tests to if statements). Changes should be accompanied by tests showing when they are necessary.

jackbsteinberg:
I removed the code that you commented on which made changes beyond just moving code around - some of them were already taken care of in the previous Window fns / get-sets PR as well

lib/mods/replace-get-set.js
```diff
@@ -24,47 +24,62 @@ function isDataProperty(type, propertyName) {
...
  * @param {Node} node - a ts-morph AST Node
  * @param {Object} imports - the current list of imports
  */
 function replaceGlobalGetSetCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      !nodeTypes.isCallExpression(node.getParent()) &&
-      isGlobal(node.getExpression())) {
-    const propertyName = node.getSymbol().getEscapedName();
-    const baseTypeName = node.getExpression().getType().getSymbol().getName();
+  let args; let type; let nodeToConsider;
```
domenic: nodeToConsider -> propertyAccessNode, maybe

lib/mods/replace-get-set.js
```diff
@@ -24,47 +24,62 @@ function isDataProperty(type, propertyName) {
...
-    } else {
-      type = 'get';
-      nodeToReplace = node;
-    }
+  if (isDataProperty(moduleName, propertyName)) {
+    return;
+  }
 
-    const original = getOriginal(propertyName, type, baseTypeName);
+  const baseObject = nodeToConsider.getExpression().getSymbol().getName();
```
domenic: Does the following tweak work?

```js
  const moduleName = nodeToConsider.getExpression().getType().getSymbol().getName();
  const baseObject = nodeToConsider.getExpression().getText();
  const propertyName = nodeToConsider.getName();
```

In particular I'm not sure why we need to indirect through getSymbol for baseObject (since baseObject is an identifier, so its text should work well), and I think .getName() on property access expressions is just a shorter version.

jackbsteinberg: Yeah that works

lib/mods/replace-get-set.js
jackbsteinberg: The name nodeToConsider was meant to be somewhat forward thinking, since when I add window gets/sets they won't be property accesses, but I can change it for now

lib/mods/replace-get-set.js
```diff
@@ -24,47 +24,62 @@ function isDataProperty(type, propertyName) {
...
  * @param {Node} node - a ts-morph AST Node
  * @param {Object} imports - the current list of imports
  */
 function replaceGlobalGetSetCalls(node, imports) {
-  if (nodeTypes.isPropertyAccessExpression(node) &&
-      !nodeTypes.isCallExpression(node.getParent()) &&
-      isGlobal(node.getExpression())) {
-    const propertyName = node.getSymbol().getEscapedName();
-    const baseTypeName = node.getExpression().getType().getSymbol().getName();
+  let args; let type; let propertyAccessNode;
```
fergald: Should be 1 statement per line.

fergald:
On Wed, 21 Aug 2019 at 05:06, Jack Steinberg <notifications@github.com>
wrote:

> After rebasing and cleaning out things that belong in a general refactor,
> I realize this PR really just serves as a general refactor to the
> replace-get-set.js file. I think instead of reparenting in this PR and
> creating another PR to do the refactor I did here, I can pivot this with
> the purpose of refactoring replace-get-set.js - thoughts?
>

I'm not sure exactly what you mean but I'm all for having 2 PRs that one
clear thing vs 1 PR that does a global refactor _and_ a local refactor.

PR #45, opened 2019-08-20
Refactor for clarity and extensibility

jackbsteinberg:


lib/imports.js
```diff
@@ -1,6 +1,30 @@
...
+const {
+  getOriginal,
+  formatReplacement,
+} = require('./originals-helper.js');
+
+/**
+ * Replaces the node in question with generated text using the
+ * original, and updates imports to get those originals.
+ *
+ * @param {Node} node - the node to replace
```
domenic: It'd be good to do mod -> moduleName throughout this file too

lib/mods/replace-methods.js
```diff
@@ -65,42 +65,42 @@ function getArgumentString(functionCall) {
...
     if (nodeTypes.isPropertyAccessExpression(node.getExpression())) {
       const propertyAccess = node.getExpression();
-      name = propertyAccess.getSymbol().getEscapedName();
+      propertyName = propertyAccess.getName();
       type = 'method';
 
       let baseType = propertyAccess.getExpression().getType();
       const baseConstructSignature = baseType.getConstructSignatures()[0];
 
+      // the method is static if it has a constructSignature
```
domenic: If the _base_ has a construct signature, right? (I.e. it is a constructor)

lib/mods/replace-methods.js
```diff
@@ -65,42 +65,42 @@ function getArgumentString(functionCall) {
...
       if (baseConstructSignature) {
         baseType = baseConstructSignature.getReturnType();
         type = 'static';
       }
 
-      mod = baseType.getSymbol().getEscapedName();
+      moduleName = baseType.getSymbol().getName();
       args = getArgumentString(propertyAccess.getParent());
```
domenic:
```suggestion
      args = getArgumentString(node);
```

lib/mods/replace-methods.js
```diff
@@ -65,42 +65,42 @@ function getArgumentString(functionCall) {
...
       callee = propertyAccess.getExpression().getSymbol().getName();
```
domenic: Elsewhere you called "callee" "baseObject", which seems better to me. And used .getText() instead of .getSymbol().getName().

lib/mods/replace-methods.js
```diff
@@ -65,42 +65,42 @@ function getArgumentString(functionCall) {
...
-        callee = mod;
+        callee = moduleName;
       }
     } else { // window functions
       type = 'window';
-      mod = 'Window';
-      name = node.getExpression().getSymbol().getEscapedName();
-      callee = mod;
+      moduleName = 'Window';
+      propertyName = node.getExpression().getSymbol().getEscapedName();
```
domenic: I think this can be `node.getExpression().getText()`

lib/mods/replace-methods.js
domenic: Also, group the `const baseConstructSignature` with this block, after the comment, to keep all things related to this test together.

lib/mods/replace-methods.js
domenic: This line can move outside the ifs, to the bottom. In fact, maybe it can move inside `replaceNodeAndUpdateImports`?

lib/mods/replace-methods.js
```diff
@@ -65,42 +65,42 @@ function getArgumentString(functionCall) {
...
+      propertyName = propertyAccess.getName();
       type = 'method';
 
       let baseType = propertyAccess.getExpression().getType();
       const baseConstructSignature = baseType.getConstructSignatures()[0];
 
+      // the method is static if it has a constructSignature
       if (baseConstructSignature) {
         baseType = baseConstructSignature.getReturnType();
         type = 'static';
```
domenic: Move the `type` assignment to be the first line inside the `if` block, for symmetry, since that's what you do in the other two cases. (And since it's kind of the "most important" thing to set.)

lib/__testfixtures__/replace-constructors.input.js
```diff
@@ -1,3 +1,9 @@
 const n = new Number();
 const t = new TypeError();
 const s = new Set([1, 2, 3]);
+class NotGlobal {
```
domenic: What is this added for? Does it fail before these changes?

lib/mods/replace-methods.js
fergald: Should we be used EscapedName everywhere? If not, something should explain why we use it sometimes and not others.

lib/__testfixtures__/replace-constructors.input.js
jackbsteinberg: Yeah I realized that it was a necessary test while refactoring and just threw it in, but it probably ought to go in a different PR

lib/mods/replace-methods.js
jackbsteinberg: I'm gonna keep it out of `replaceNodeAndUpdateImports` for now because it only ever gets used in the method replacement mod, and args for the get-set mod are done in a different way. There is maybe a case to be made to make `getArgumentString` into a more generic method with a switch statement like some of the other methods, but I don't think it's necessary at the moment.

PR #46, opened 2019-08-21
Add mod for replacing non-property access window get-sets

jackbsteinberg:


lib/__testfixtures__/window.input.js
```diff
@@ -4,3 +4,5 @@ setTimeout(() => {
 const f = parseFloat('5.5');
 const a = window.closed;
 window.status = 'open';
+status = 'closed';
+const s = status;
```
domenic: Can you test the getter case in trickier positions? E.g. passing to a function, or `navigator.getBattery()` (should translate to Window_navigator_get).

lib/mods/replace-methods.js
```diff
@@ -73,7 +73,7 @@ function replaceGlobalMethodCalls(node, imports) {
     // property access methods
     if (nodeTypes.isPropertyAccessExpression(node.getExpression())) {
       const propertyAccess = node.getExpression();
-      propertyName = propertyAccess.getName();
+      propertyName = propertyAccess.getName(); // getSymbol.getEscapedName
```
domenic: Is this change intentional?

lib/__testfixtures__/window.output.js
```diff
@@ -1,7 +1,9 @@
-import { setTimeout as Window_setTimeout, parseFloat as Window_parseFloat, closed_get as Window_closed_get, status_set as Window_status_set } from "std:global/Window";
+import { setTimeout as Window_setTimeout, parseFloat as Window_parseFloat, closed_get as Window_closed_get, status_set as Window_status_set, status_get as Window_status_get } from "std:global/Window";
 import { apply as Reflect_apply } from "std:global/Reflect";
 Window_setTimeout(() => {
 }, 3000);
 const f = Window_parseFloat('5.5');
 const a = Reflect_apply(Window_closed_get, window);
 Reflect_apply(Window_status_set, window, ['open']);
+Reflect_apply(Window_status_set, window, ['closed']);
+const s = Reflect_apply(Window_status_get, window);
```
domenic: OK, so seeing these results made me realize something. `window` is itself a getter, so we don't want it to appear in the output. Instead we want to do `Window_closed_get()` and `Window_status_set('closed')`. This is a very special case. Note that this only works because of https://heycam.github.io/webidl/#dfn-attribute-getter 1.1.2.1 or https://heycam.github.io/webidl/#dfn-attribute-setter 1.4.5.1.

We can either do that as part of this PR, or as a separate one on top. If a separate one, then let's create a new issue to track.

Note that this should apply to both the "bare" case and the non-bare case.

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,60 @@ function isDataProperty(type, propertyName) {
...
-function isGlobalSet(node) {
+function isSet(node) {
   return nodeTypes.isBinaryExpression(node) &&
-    node.getOperatorToken().getText() === '=' &&
-    isGlobalGet(node.getLeft());
+    node.getOperatorToken().getText() === '=';
+}
+
+/**
+ * @param {Node} node - the ts-morph AST Node to evaluate
```
domenic: Can you state what type of node it is? I think it's Identifier?

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,60 @@ function isDataProperty(type, propertyName) {
...
+ */
+function isWindowGlobal(node) {
+  // filter constructors
+  const baseConstructSignature = node.getType().getConstructSignatures()[0];
+  if (baseConstructSignature) {
+    return false;
+  }
+
+  // filter namespaces
+  if (node.getType && node.getType().getText() === node.getText()) {
```
domenic: When is `node.getType` falsy? Wouldn't that cause a crash on line 55?

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,60 @@ function isDataProperty(type, propertyName) {
...
+    return false;
+  }
+
+  const parent = node.getParent();
+  // filter data properties
+  const propertyName = node.getText();
+  if (parent.getExpression && parent.getExpression().getType() &&
+      isDataProperty(parent.getExpression().getType().getSymbol() &&
+      parent.getExpression().getType().getSymbol().getName(),
+      propertyName)) {
```
domenic: This if repeats `parent.getExpression().getType()` several times.

In what cases is `parent.getExpression().getType()` falsy? In what cases is `parent.getExpression().getType().getSymbol()` falsy?

Avoid checking for `parent.getExpression`. Instead check the parent node type. I believe you want to check if it's a property access node?

I think the ideal code here would be

```js
const baseName = parent.getExpression().getType().getSymbol().getName();
```

inside an if that tests if it's a property access expression. But it's possible getType or getSymbol could fail. You need to comment on what cases that happens, though.

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,60 @@ function isDataProperty(type, propertyName) {
...
+    return false;
+  }
+
+  // check for global
+  const symbol = node.getSymbol();
+  if (symbol && symbol.getDeclarations()[0]) {
```
domenic: This would be cleaner with

```js
const declaration = symbol && symbol.getDeclarations[0];
```

lib/mods/replace-get-set.js
domenic: In general I'm having a hard time understanding which case this is meant to cover, exactly. I suggest for all the filters, you add an example. E.g.

```js
// Filter constructors. Example: TypeError
// Filter namespaces. Example: CSS
// Filter data properties. Example: ...??
```

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,60 @@ function isDataProperty(type, propertyName) {
...
+  // filter data properties
+  const propertyName = node.getText();
+  if (parent.getExpression && parent.getExpression().getType() &&
+      isDataProperty(parent.getExpression().getType().getSymbol() &&
+      parent.getExpression().getType().getSymbol().getName(),
+      propertyName)) {
+    return false;
+  }
+
+  // check for global
```
domenic: I think moving this check first in the function would be good.

lib/mods/replace-get-set.js
```diff
@@ -55,28 +90,50 @@ function isGlobalSet(node) {
...
     type = 'set';
+    nodeToReplace = node.getLeft();
     args = node.getRight().getText();
-  } else {
+  } else if (isWindowGlobal(node)) { // non-property access window get-set
+    const parent = node.getParent();
+
+    if (isSet(parent)) { // non-property access window set
+      type = 'set';
+      nodeToReplace = parent;
```
domenic: Don't replace parents! Needs a bit of restructuring to be more like the global get/global set cases.

In particular I think the structure that would work is:

- if isGlobalPropertyAccess
- else if isSet && isGlobalPropertyAccess(left)
- else if isWindowGlobal
- else if isSet && isWindowGlobal(left)

Ideally this should allow you to go back to having a single replaceNodeAndUpdateImports call.

lib/mods/replace-methods.js
jackbsteinberg: No, good catch

lib/__testfixtures__/window.output.js
jackbsteinberg: I set up an #49 to track this, but I think it can relatively simply be added to this PR

lib/mods/replace-get-set.js
jackbsteinberg: I don't think we know that for certain, it will only pass through if it's an Identifier, but as of right now it gets every node. Looking at this with that perspective though, I'm going to add a check in here for Identifiers.

lib/__testfixtures__/window.output.js
jackbsteinberg: Going to fix this in another PR, I just looked into it and it's pretty simple to add for the bare case, but more complicated for the non-bare case. Worth splitting into its own CL

jackbsteinberg: Never mind 😅, adding this turned out to be relatively simple, and simplified some of the rest of the logic as well

lib/mods/replace-get-set.js
jackbsteinberg: Unless there are any `window` data properties I can simplify this by checking if the parent is a property access expression and just exiting out if so (no need to check `isDataProperty`)

jackbsteinberg: I'll go through and add some explanations on my filters and conditionals - it's likely some of them can be simplified

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,70 @@ function isDataProperty(type, propertyName) {
...
+ * identifier, by eliminating other possible types of globals
+ * the identifier could be
+ *
+ * @param {Node} node - the ts-morph AST Node to evaluate
+ *
+ * @return {boolean} whether the node is a window global
+ */
+function isWindowGlobal(node) {
+  // filter non-globals
+  const symbol = node.getSymbol();
```
domenic: Can we dedupe this code with isGlobal from originals-helper.js?

lib/mods/replace-get-set.js
domenic: I am pretty sure it is an identifier. Can you give an example (e.g. by console.loging it) where it is not?

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,70 @@ function isDataProperty(type, propertyName) {
...
+  const declaration = symbol && symbol.getDeclarations()[0];
+  if (!declaration ||
+    !declaration.getSourceFile().getFilePath()
+        .includes('/node_modules/typescript')) {
+    return false;
+  }
+
+  // filter constructors - e.g. Document
+  const baseConstructSignature = node.getType().getConstructSignatures()[0];
```
domenic: There is no "base" involved, since this is not a property access expression.

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,70 @@ function isDataProperty(type, propertyName) {
...
+
+  // filter namespaces - e.g. CSS
+  if (node.getType().getText() === node.getText()) {
+    return false;
+  }
+
+  // filter data properties - e.g. array.length
+  const parent = node.getParent();
+  const propertyName = node.getText();
+  let baseName;
```
domenic: Why declare this outside the if?

lib/mods/replace-get-set.js
```diff
@@ -55,29 +100,53 @@ function isGlobalSet(node) {
 function replaceGlobalGetSetCalls(node, imports) {
```
domenic: The flow if this function is now very clean; nice.

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,70 @@ function isDataProperty(type, propertyName) {
...
+  if (node.getType().getText() === node.getText()) {
+    return false;
+  }
+
+  // filter data properties - e.g. array.length
+  const parent = node.getParent();
+  const propertyName = node.getText();
+  let baseName;
+
+  if (nodeTypes.isPropertyAccessExpression(parent)) {
```
domenic: Why do we need this check given the check on line 145? What test fails if you remove it?

lib/mods/replace-get-set.js
jackbsteinberg: Attached is the output of some of this, because it's checking every node. What I can do is add a first filter checking if it's an identifier, but unless I did that in both if statements using `isWindowGlobal` then it's still gonna take in `Node`

```    
    const a = window.closed;
    const a = window.closed
    a = window.closed
    a
    window.status = 'open';
    status = 'closed';
    status = 'closed'
    status
    const s = status;
    const s = status
    s = status
    s
    status
    const bat = navigator.getBattery();
    const bat = navigator.getBattery()
    bat = navigator.getBattery()
    bat
    navigator.getBattery()
    navigator.getBattery
    navigator
    getBattery
    Window_alert(status);
    Window_alert(status)
    Window_alert
    status
```

domenic: I see. Well yeah, the rest of the code in this function doesn't make a lot of sense for non-identifiers. So maybe just add `if (!identifier) { return; }` at the top of this function?

lib/mods/replace-get-set.js
jackbsteinberg: I looked into this we actually can't trivially, since `isGlobal` is checking if the type of the node is declared in TypeScript, whereas for window we want to see if the symbol itself was declared in TypeScript

lib/mods/replace-get-set.js
jackbsteinberg: It fails with an error replacing tree in get-set and property-access, the culprit being an array.length which seems to mess things up.

domenic:
OK, after our in-person discussion to try to help me understand the data property check, here is what I wrote up. I think we should include it as a large comment block in the source on top of the function in question:

---

Checks if `node` is an identifier that is a property of a global, but not part of a property access expression (as those will be caught earlier). For example, `status = 'foo'` or `func(name)`, since `status` and `name` are properties of the global.

There are two possible approaches to achieving this:

1. Look at the node's TypeScript symbol, and check that it is a variable declaration that comes from the built-in TypeScript definitions. Variable declarations appears to be how TypeScript models global properties; see for example https://github.com/microsoft/TypeScript/blob/4bddf553282c9d5b76e5fb785d7c6e74e67f034a/lib/lib.dom.d.ts#L19571. This requires then eliminating namespaces and constructors, which are also modeled with variable declarations.

2. Start with the assumption that all nodes whose symbols come from the built-in TypeScript types might be global variables, then eliminate the cases that we know are not. In particular, after eliminating namespaces and constructors, we are left with property access on built-in types (e.g. `someNode.localName` or `someArray.length`). All of these will already have been replaced with their get-originals counterparts, so we mostly don't have to worry---except for data properties, which we left alone. So just be sure to skip data properties too.

We choose approach (2) here for now, as it seems likely that in any cases where they disagree, (2) will cause more rewriting than (1). This is the preferable failure mode, as the rewriting will be obviously wrong and cause runtime failures in the rewritten code, whereas failing to rewrite means the rewritten code is vulnerable to patched globals.

At this time we know of no examples that give different results for (1) or (2); indeed we know of no examples where (2) gives the incorrect answer. If we find some, we can reassess, and either tweak our approach to (2) or switch to (1), as seems appropriate.

---

lib/mods/replace-get-set.js
```diff
@@ -25,25 +25,74 @@ function isDataProperty(type, propertyName) {
...
+  }
+
+  // filter data properties - e.g. array.length
+  const parent = node.getParent();
+  if (nodeTypes.isPropertyAccessExpression(parent)) {
+    const propertyName = node.getText();
+    const baseName = parent.getExpression().getType().getSymbol().getName();
+
+    if (isDataProperty(baseName, propertyName)) {
+      console.log('caught:', node.getParent().getText());
```
domenic: Remove this

PR #47, opened 2019-08-21
Remove IDL checking for isGlobal

jackbsteinberg:


lib/originals-helper.js
```diff
@@ -1,13 +1,12 @@
...
 /**
  * @param {Node} node - a ts-morph AST Node
  *
- * @return {boolean} whether the type of the node is from TypeScript definitions
+ * @return {boolean} whether the type of the node is global
+ * based on TypeScript definitions
  */
-function nodeInTSDefs(node) {
+function isGlobal(node) {
   const symbol = node.getType().compilerType.symbol;
```
domenic: It occurs to me that this could probably be `node.getType().getSymbol()`. A minor thing to maybe fix in a follow-up pull request.

jackbsteinberg: For some reason, not quite sure why, that doesn't work

PR #48, opened 2019-08-21
Add extra testing to fill coverage

jackbsteinberg:


lib/__testfixtures__/property-access.output.js
```diff
@@ -3,3 +3,9 @@ import { apply as Reflect_apply } from "std:global/Reflect";
 const a = [1, 2, 3];
 const l = a.length;
 const str = Reflect_apply(Array_toString, a, []);
+const obj = {
```
domenic: Maybe name it `notGlobal` or something so it's obvious what you're testing.

PR #50, opened 2019-08-22
Remove jscodeshift dependency and usage

### LINKED ISSUE #30: Remove jscodeshift dependency

jackbsteinberg:
Resolves #30


lib/__tests__/test.js
```diff
@@ -2,7 +2,45 @@
...
+    });
+  });
+}
+
+/**
+ * @param {string} dirname - the testing directory
+ * @param {string} transformName - the name of the transform
+ * @param {string} filePrefix - the test file prefix
+ */
+function runTest(dirname, transformName, filePrefix) {
```
domenic: Let's make this a bit less generic by removing transformName. Then we can just require() the rewriter at the top of the file. We don't need infrastructure to test multiple different transforms.

lib/__tests__/test.js
```diff
@@ -2,7 +2,45 @@
...
+  const fixtureDir = path.join(dirname, '..', '__testfixtures__');
+  const inputPath = path.join(fixtureDir, filePrefix + '.input.js');
+  const source = fs.readFileSync(inputPath, 'utf8');
+  const expectedOutput = fs.readFileSync(
+      path.join(fixtureDir, filePrefix + '.output.js'),
```
domenic: Might as well have an `outputPath` for symmetry with `inputPath`, even if you only use it once.

lib/__tests__/test.js
```diff
@@ -2,7 +2,45 @@
...
+  );
+
+  // Assumes transform is one level up from __tests__ directory
+  const transform = require(path.join(dirname, '..', transformName));
+  const rewriterOutput = transform({
+    path: inputPath,
+    source,
+  });
+
+  expect((rewriterOutput || '').trim()).toEqual(expectedOutput.trim());
```
domenic: Does this work without `trim()` or the `|| ''`? That'd be cleaner if we can manage it.

lib/index.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const fs = require('fs');
+const rewriter = require('./rewriter');
+
+/**
+ * @param {string} filePath - the path to the file
+ */
+function rewriteFile(filePath) {
+  const file = {
```
domenic: Here I feel like the `file` variable doesn't add much, so we can just change it to `rewriter({ ... })`.

lib/index.js
```diff
@@ -0,0 +1,24 @@
...
+    path: filePath,
+    source: fs.readFileSync(inputPath, 'utf8'),
+  };
+  rewriter(file);
+}
+
+// rewriter execution
+const inputPath = process.argv[2];
+console.log(inputPath);
```
domenic: We probably don't want to commit this, or if we do, it'd be something more like `Rewriting ${inputPath}` inside `rewriteFile`.

I think on balance it's better to not print anything when your program succeeds. I will try to find a source for that "rule" somewhere tomorrow...

README.md
```diff
@@ -27,37 +27,6 @@ where `PATH` is a file or directory of files to be rewritten.
 
 The Get-Originals Rewriter provides the following options:
```
domenic: This sentence is now kind of dangling. You want to mention the usage as either file or directory.

lib/__tests__/test.js
jackbsteinberg: The `|| ''` I can take out no problem but the testing is getting mad at me about having trailing spaces in my files. Alternatives are remove the trailing spaces, or remove the trim.

README.md
jackbsteinberg: I just removed it, since I address that file / directory usage in the sentence before. The README is in major need of an update generally, I think I'll have a PR for that next week.

lib/index.js
fergald: I don't know a source for it but I agree. If every tool announced success then scripts that combine tools would be full of success output and important warnings in the output would be hard to find.

lib/__tests__/test.js
```diff
@@ -2,7 +2,41 @@
...
+ * @param {string} dirname - the testing directory
+ * @param {string} filePrefix - the test file prefix
+ */
+function runTest(dirname, filePrefix) {
+  const fixtureDir = path.join(dirname, '..', '__testfixtures__');
+  const inputPath = path.join(fixtureDir, filePrefix + '.input.js');
+  const outputPath = path.join(fixtureDir, filePrefix + '.output.js');
+  const source = fs.readFileSync(inputPath, 'utf8');
+  const expectedOutput = fs.readFileSync(outputPath, 'utf8'
+  );
```
fergald: spurious newline here.

lib/__tests__/test.js
domenic: I'm a bit torn here, but it's probably cleaner to remove the trailing spaces. We should ideally be testing against the exact output, and having our fixture files represent the exact output.

This may make editing the files tricky if you (like me) have your editor configured to automatically add trailing newlines to files. The best way to get around this is with [EditorConfig](https://editorconfig.org/). In particular, install the [VSCode EditorConfig plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig), and then add (either in this PR or a separate one) an .editorconfig file.

A good base .editorconfig file to use is something like https://github.com/domenic/worm-scraper/blob/master/.editorconfig . I.e., set reasonable defaults for the whole project, then have exceptions for specific directories.

Once this is all setup, VS Code---and many other editors---would read "whether to add a final newline or not" from the .editorconfig file, and thus do the right thing, adding it for most files, but not messing with the fixture outputs.

README.md
```diff
@@ -25,39 +25,6 @@ npm run rewrite PATH
 
 where `PATH` is a file or directory of files to be rewritten.
```
domenic: Have you tested that this works, or do you need `npm run rewrite -- PATH`? I'm never sure when you need the `--` with `npm run` or not.

lib/index.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
domenic: Traditionally we place the "executable" files in a bin/ folder instead of lib/. Also, it's good to add a shebang to them as the first line: `#!/usr/bin/env node`. This makes them directly executable, using `./bin/index.js`.

package.json
```diff
@@ -4,7 +4,6 @@
   "description": "Rewriter presubmit script for get-originals work",
   "main": "lib/index.js",
```
domenic: This is not a big deal since we're not really (yet??) creating a distributable npm package, but the best practice here would be to make the `main` lib/rewriter.js, as the main "programmatic" entry point people would get from `require("get-originals-rewriter")`, and then add a `"bin": "bin/index.js"`, as the executable.

README.md
domenic: Actually I wonder if it'd be simpler to just tell people to do `node bin/index.js`, and then omit the `rewrite` script in package.json entirely. Now that we don't have the jscodeshift indirection, using the npm script stuff becomes much less necessary.

jackbsteinberg: It works as `npm run rewrite PATH` at the moment. We could tell people to do either, I don't know what the convention with using `npm run` is

domenic: I think removing the rewrite script makes sense at this point.

jackbsteinberg: Ok I can take it out, should I change the name to `rewrite.js` or something more semantic?

domenic: Yeah that's probably a good idea.

lib/__tests__/test.js
```diff
@@ -11,26 +44,11 @@ function runTestFixtures() {
   walkDir(path.join(__dirname, '../__testfixtures__'), (file) => {
     const [filename, ioType] = file.split('.');
```
domenic: Since it's not an actual filename, I think filePrefix is better here.

lib/__tests__/test.js
```diff
@@ -2,7 +2,40 @@
...
+      runTest(dirname, filePrefix);
+    });
+  });
+}
+
+/**
+ * @param {string} dirname - the testing directory
+ * @param {string} filePrefix - the test file prefix
+ */
+function runTest(dirname, filePrefix) {
```
domenic: It'd be better if this function only took a filePrefix. To do that, you could replace the first line with

```js
path.resolve(filePrefix, '..', '__testfixtures')
```

lib/execution-helper.js
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Recursively walks through a folder and its children,
+ * calling a callback on each file
+ *
+ * @param {String} dir - directory to walk through
+ * @param {Function} callback - operation to perform on each file
+ */
+function walkDir(dir, callback) {
+  fs.readdirSync(dir).forEach((f) => {
+    const dirPath = path.join(dir, f);
```
domenic: This is not necessarily a directory, so I think `directoryEntryPath` is probably better

lib/execution-helper.js
```diff
@@ -0,0 +1,23 @@
...
+    const isDirectory = fs.statSync(dirPath).isDirectory();
+    isDirectory ? walkDir(dirPath, callback) : callback(f);
```
domenic: I think this would read better as a normal if statement, instead of the ternary.

lib/__tests__/test.js
```diff
@@ -2,7 +2,40 @@
...
+  const outputPath = path.join(fixtureDir, filePrefix + '.output.js');
+  const source = fs.readFileSync(inputPath, 'utf8');
+  const expectedOutput = fs.readFileSync(outputPath, 'utf8');
+
+  const rewriterOutput = rewriter({
+    path: inputPath,
+    source,
+  });
+
+  expect(rewriterOutput.trim()).toEqual(expectedOutput.trim());
```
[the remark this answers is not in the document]
jackbsteinberg: I'm not 100% sure what strategy you want to take with the trimming. I don't currently have anything that auto-adds it I've been doing it myself, should I just take the trailing spaces out? It gives less consistent code but the alternative is trimming.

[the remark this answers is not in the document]
domenic: Yes, I think the .output.js files should match exactly what we output. So yeah, just take the trailing spaces out then.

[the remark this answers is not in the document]
jackbsteinberg: Update: needed to remove both trims

README.md
fergald: You only need `--` if your script is taking flags. Without the `--` the flags would be parsed by npm's flag parser. So since rewrite.js doesn't do any flag parsing (yet) the `--` are not needed.

lib/__tests__/test.js
fergald: Can we just make the rewriter put a newline at the end?

PR #53, opened 2019-08-23
Fix rewriting files and add integration test

### LINKED ISSUE #40: npm run rewrite fails
Every attempt to run this on a very simple input file fails. It seems like it was broken by e35be3c94da44f230c6cf155d462b8380913ec05.

```
$@ npm run rewrite test2.js

> get-originals-rewriter@1.0.0-pre rewrite /home/fergal/github/jackbsteinberg/get-originals-rewriter
> jscodeshift -t ./lib/rewriter.js "test2.js"

Processing 1 files... 
Spawning 1 workers...
Sending 1 files to free worker...
/home/fergal/github/jackbsteinberg/get-originals-rewriter/node_modules/ts-morph/dist/codeBlockWriter/code-block-writer.js:13
tempThis["CodeBlockWriter"] = code_block_writer_1.default;
                            ^

TypeError: Cannot set property 'CodeBlockWriter' of undefined
    at Object.<anonymous> (/home/fergal/github/jackbsteinberg/get-originals-rewriter/node_modules/ts-morph/dist/codeBlockWriter/code-block-writer.js:8:1)
    at Module._compile (internal/modules/cjs/loader.js:701:30)
    at Module._compile (/home/fergal/github/jackbsteinberg/get-originals-rewriter/node_modules/pirates/lib/index.js:99:24)
    at Module._extensions..js (internal/modules/cjs/loader.js:712:10)
    at Object.newLoader [as .js] (/home/fergal/github/jackbsteinberg/get-originals-rewriter/node_modules/pirates/lib/index.js:104:7)
    at Module.load (internal/modules/cjs/loader.js:600:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:539:12)
    at Function.Module._load (internal/modules/cjs/loader.js:531:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
All done. 
Results: 
0 errors
0 unmodified
0 skipped
0 ok
Time elapsed: 1.338seconds 
```

### LINKED ISSUE #52: Rewriter doesn't actually rewrite
At this time the rewriter when run processes the file, determining what the rewritten output would be, but doesn't use that output to actually rewrite the file.

The code needs to be updated to reflect the requirement of actually rewriting, and should get an integration test for this at the same time.

jackbsteinberg:
Resolves #52 
Resolves #40 (again)

### DIFF
--- a/bin/rewrite.js
+++ b/bin/rewrite.js
@@ -8,10 +8,12 @@ const rewriter = require('../lib/rewriter');
  * @param {string} filePath - the path to the file
  */
 function rewriteFile(filePath) {
-  rewriter({
+  const output = rewriter({
     path: filePath,
     source: fs.readFileSync(inputPath, 'utf8'),
   });
+
+  fs.writeFileSync(filePath, output);
 }
 
 // rewriter execution
--- a/lib/__tests__/test.js
+++ b/lib/__tests__/test.js
@@ -4,6 +4,36 @@ const fs = require('fs');
 const path = require('path');
 const {walkDir} = require('../execution-helper');
 const rewriter = require('../rewriter');
+const {exec} = require('child_process');
+
+const integrationInput = fs.readFileSync(
+    './lib/__testfixtures__/static-methods.input.mjs', 'utf8'
+);
+const integrationOutput = fs.readFileSync(
+    './lib/__testfixtures__/static-methods.output.mjs', 'utf8'
+);
+
+/**
+ * Runs a test on the rewriter script's file-overwrite
+ */
+async function runIntegrationTest() {
+  const path = './integration.js';
+  // write input to file
+  fs.writeFileSync(path, integrationInput, 'utf8');
+
+  // rewrite file
+  await exec(`./bin/rewrite.js ${path}`);
+
+  const fileOutput = fs.readFileSync(path, 'utf8');
+
+  // check against expected output
+  test('the rewriter command works as expected', () => {
+    expect(fileOutput).toEqual(integrationOutput);
+  });
+
+  // delete file
+  fs.unlinkSync(path);
+}
 
 /**
  * @param {string} filePrefix - the test file prefix
@@ -48,3 +78,4 @@ function runTestFixtures() {
 jest.autoMockOff();
 
 runTestFixtures();
+runIntegrationTest();
--- a/lib/__tests__/test.js
+++ b/lib/__tests__/test.js
@@ -23,11 +23,10 @@ async function runIntegrationTest() {
 
   // rewrite file
   await exec(`./bin/rewrite.js ${path}`);
-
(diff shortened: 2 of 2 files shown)


lib/__tests__/test.js
```diff
@@ -4,6 +4,36 @@ const fs = require('fs');
...
+/**
+ * Runs a test on the rewriter script's file-overwrite
+ */
+async function runIntegrationTest() {
+  const path = './integration.js';
+  // write input to file
+  fs.writeFileSync(path, integrationInput, 'utf8');
+
+  // rewrite file
+  await exec(`./bin/rewrite.js ${path}`);
```
domenic: This doesn't return a promise, so I'm not sure how this works... I think you want execSync instead.

lib/__tests__/test.js
```diff
@@ -4,6 +4,36 @@ const fs = require('fs');
...
+);
+const integrationOutput = fs.readFileSync(
+    './lib/__testfixtures__/static-methods.output.mjs', 'utf8'
+);
+
+/**
+ * Runs a test on the rewriter script's file-overwrite
+ */
+async function runIntegrationTest() {
+  const path = './integration.js';
```
domenic: All the work should be inside the `test(() => { ... })` block so that any failures get reported appropriately.

lib/__tests__/test.js
```diff
@@ -4,6 +4,36 @@ const fs = require('fs');
...
+  // write input to file
```
domenic: All three of the comments in this file are kind of redundant with the code; I'd remove them.

lib/__tests__/test.js
```diff
@@ -4,6 +4,36 @@ const fs = require('fs');
...
+
+  const fileOutput = fs.readFileSync(path, 'utf8');
+
+  // check against expected output
+  test('the rewriter command works as expected', () => {
+    expect(fileOutput).toEqual(integrationOutput);
+  });
+
+  // delete file
+  fs.unlinkSync(path);
```
domenic: This should be in a `finally {}` block so that even if the test fails it still gets deleted.

jackbsteinberg:
I knew it seemed too easy. This update should resolve that.

bin/rewrite.js
```diff
@@ -8,10 +8,12 @@ const rewriter = require('../lib/rewriter');
  * @param {string} filePath - the path to the file
  */
 function rewriteFile(filePath) {
-  rewriter({
+  const output = rewriter({
     path: filePath,
     source: fs.readFileSync(inputPath, 'utf8'),
```
fergald: inputPath does not exist but if this was supposed to be filePath then I think it's still wrong, I don't think users will want their original source code over-written. I've re-opened #52

PR #56, opened 2019-08-26
Add improved data property detection

### LINKED ISSUE #33: Solid detection of data properties
As discussed in #26 and #28, we'll initially land accessor replacement with a list that prevents transforming string length and array length. We should expand this to be a solid list.

I don't know of any good way to do this besides just having a list of all data properties that exist in the platform. So I will make such a list. Then we should update the tests and implementation to ensure they don't get rewritten.

jackbsteinberg:
Resolves #33 
Adds extra rules for data properties and tests each (function instances included)


lib/mods/replace-get-set.js
```diff
@@ -122,10 +132,12 @@ function isWindowGlobal(node) {
   const parent = node.getParent();
   if (nodeTypes.isPropertyAccessExpression(parent)) {
     const propertyName = node.getText();
-    const baseName = parent.getExpression().getType().getSymbol().getName();
+    if (symbol.compilerSymbol.parent) {
```
fergald: a comment on what case this is catching would be helpful.

PR #60, opened 2019-08-27
Refactors + improve detection of data properties

domenic:
As discussed offline while Jack shifts focus to work on a design doc I wanted to finish up #33/#56. I did so by first doing a couple small refactorings, which made it easier for me to, in the last commit, improve our data property detection.

It might be easiest to review the commits one by one, although GitHub only lets you leave comments on the overall diff.


domenic:
@fergal your review would still be appreciated if you have time, but I'll merge this for now.

lib/originals-helper.js
```diff
@@ -15,6 +15,29 @@ function isGlobal(node) {
...
+ * @param {Type} type - a ts-morph type
+ *
+ * @return {string} the get-originals module name corresponding to the
+ * type. (This is sometimes different from the TypeScript type name.)
+ */
+function getModuleNameForType(type) {
+  // `type.getText()` does not work, because it returns things like
+  // "Promise<any>" or "any[]". The name of the type's symbol is what we want:
+  // "Promise" or "Array".
+  return type.getSymbol().getName();
```
fergald: Is there something more available that lets your distinguish the builtin Array from a random class named Array?

lib/mods/replace-get-set.js
```diff
@@ -119,15 +130,14 @@ function isWindowGlobal(node) {
     return false;
   }
 
-  // filter data properties - e.g. array.length
+  // Filter property access where this Identifier is the property, e.g.
```
fergald: Nit, s/Filter/Exclude/ as filter may be positive or negative.

lib/mods/replace-methods.js
```diff
@@ -32,6 +32,10 @@ function isPropertyOfNamespace(propertyAccess) {
   const baseModuleName = getModuleNameForNode(base);
 
   const baseSymbol = base.getSymbol();
+  if (!baseSymbol) {
```
fergald: Would be good to document when this occurs and/or add unit tests for these isFoo helpers to make it clear what cases we are thinking about.

lib/originals-helper.js
domenic: Yes; that's done via the isGlobal check that we do at various locations in the code.

lib/mods/replace-methods.js
domenic: The integration tests cover this case, but yeah, documenting it would be good.

PR #67, opened 2019-09-04
Prune unnecessary constructor rewrite code

### LINKED ISSUE #62: replace-constructors mod replaces node for no reason
While describing the mods in the design document I noticed that in replace-constructors.js line 24 replaces the constructor with the original. However, the `getOriginal` function doesn't modify the node when it has type constructor, because constructors only need to be imported, not rewritten.

A bit of code can be removed for replacing the constructor Identifier with identical text and for getting the original of a constructor.

jackbsteinberg:
Resolves #62


lib/mods/import-constructors.js
```diff
@@ -4,26 +4,24 @@ const nodeTypes = require('ts-morph').TypeGuards;
...
  */
-function replaceGlobalConstructors(node, imports) {
+function importGlobalConstructors(node, imports) {
   if (nodeTypes.isNewExpression(node) &&
       isGlobal(node)) {
-    const identifier = node.getExpression();
     const moduleName = getModuleNameForNode(node);
-    const original = getOriginal(moduleName, 'constructor');
+    // original has the same name as module for constructors
+    const original = moduleName;
```
domenic: Now there is an unused case in getOriginal that checks for type = constructor. Either we should continue to use getOriginal, or we should remove it from getOriginal

jackbsteinberg: It's not actually unused, it gets used in `generateImportText` - I compromised by leaving `constructor` in `getOriginal` so I could call it at the beginning of that function since it gets used in all the other cases.

Something I could do instead is remove it from `getOriginal` and only call that function in `generateImportText` after checking if the type is `constructor`

domenic: Right, that seems more correct to me.

jackbsteinberg: Cool, I'll switch to that

lib/imports.js
```diff
@@ -36,22 +36,28 @@ function replaceNodeAndUpdateImports(node, name, type,
  * @return {string} the formatted import text
  */
 function generateImportText(identifier, type, moduleName) {
-  const original = getOriginal(identifier, type, moduleName);
+  // constructors have no getOriginal value
+  // because the original is the same as the module name
+  let original;
+  if (type !== 'constructor') {
+    original = getOriginal(identifier, type, moduleName);
+  }
```
domenic: I think it'd be cleaner to just do `if (type === 'constructor') { return moduleName; }`, before doing `const original = ...`.

jackbsteinberg: Yep that is in fact much cleaner, wish I had thought of that

PR #69, opened 2019-09-05
Add support for chained gets, sets, and methods

### LINKED ISSUE #29: Nested accessors and methods
Currently when the rewriter encounters nested accessors and methods it will only convert the outermost to its original:

```
const pathLength = url.pathname.length;
const expressionName = node.getExpression().getSymbol().getName();

---- REWRITTEN ----

const pathLength = Reflect_apply(String_length_get, url.pathName);
const expressionName = Reflect_apply(Symbol_getName, node.getExpression().getSymbol(), []);
```

This is possibly because parents of observed nodes are being rewritten, which prevents the algorithm from continuing to inspect and properly rewrite all the children.

jackbsteinberg:
Resolves #29

### DIFF
--- a/lib/mods/replace-get-set.js
+++ b/lib/mods/replace-get-set.js
@@ -47,9 +47,13 @@ function isDataPropertyAccess(node) {
  * @return {boolean} whether the node is a global get/set
  */
 function isGlobalPropertyAccess(node) {
+  // Checks if the parent is a call expression and
+  // the node is its expression (not one of its arguments)
+  const isCallExpressionChild = nodeTypes.isCallExpression(node.getParent()) &&
+      node.getParent().getExpression() === node;
+
   return nodeTypes.isPropertyAccessExpression(node) &&
-    !(nodeTypes.isCallExpression(node.getParent()) &&
-    node.getParent().getExpression() === node) &&
+    !isCallExpressionChild &&
     isGlobal(node.getExpression());
 }
 
--- a/lib/mods/replace-get-set.js
+++ b/lib/mods/replace-get-set.js
@@ -48,7 +48,8 @@ function isDataPropertyAccess(node) {
  */
 function isGlobalPropertyAccess(node) {
   return nodeTypes.isPropertyAccessExpression(node) &&
-    !nodeTypes.isCallExpression(node.getParent()) &&
+    !(nodeTypes.isCallExpression(node.getParent()) &&
+    node.getParent().getExpression().getText() !== 'Reflect_apply') &&
     isGlobal(node.getExpression());
 }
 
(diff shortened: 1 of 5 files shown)


lib/mods/replace-get-set.js
```diff
@@ -48,7 +48,8 @@ function isDataPropertyAccess(node) {
  */
 function isGlobalPropertyAccess(node) {
   return nodeTypes.isPropertyAccessExpression(node) &&
-    !nodeTypes.isCallExpression(node.getParent()) &&
+    !(nodeTypes.isCallExpression(node.getParent()) &&
+    node.getParent().getExpression() === node) &&
```
domenic: Is it possible to express what this condition is asking? e.g.

```js
const ___something___ = nodeTYpes.isCallExpression(node.getParent()) && node.getParent().getExpression === node;
```

and then just use `!___something__`? Where `__something__` is a semantic name indicating what actually you're testing for...

jackbsteinberg: I was considering documenting via comments but this seems like a good way to do it, I'll change to this

jackbsteinberg:
We only need the recursive application for methods and gets/sets, since, as of a few PRs ago, constructors don't rewrite. It's also unnecessary for the remove console mod as that only rewrites to `undefined` or deletes itself.

lib/mods/replace-get-set.js
jackbsteinberg: Updated to a separate var with a explanatory comment

fergald:
> We only need the recursive application for methods and gets/sets, since, as of a few PRs ago, constructors don't rewrite. It's also unnecessary for the remove console mod as that only rewrites to `undefined` or deletes itself.

I still haven't figured out how this change works. If recursion only happens within a module, how does

```js
document.body.querySelector('class');
```

get rewritten correctly, that seems like we have to recurse into 2 modules.

Maybe I don't understand how the tree traversal works at all.

jackbsteinberg:
I did some digging and that particular example works, but there is still an issue.

```js
document.body.querySelector('class');
```

This would first have the `replaceMethods` mod run, which would transform it into:

```js
Reflect_apply(HTMLElement_querySelector, document.body, ['class']
```

After the `replaceMethods` mod recursed and found no more methods to replace, it would have the `replaceGetsSets` mod run, which would replace `document.body` giving this:

```js
Reflect_apply(HTMLElement_querySelector, Reflect_apply(Document_body_get, document), ['class']);
```

After this `replaceGetsSets` would recurse, replacing `document` and leaving us with the final result of:

```js
Reflect_apply(HTMLElement_querySelector, Reflect_apply(Document_body_get, Window_document_get()), ['class']);
```

jackbsteinberg:
This example works, but only because of the order of the accesses, with method coming before gets/sets. If, for example, we had something closer to:

```js
const otherDiff = document.querySelector('body').textContent;
```

Then the `textContent` portion of the statement would not be rewritten, leaving:

```js
Reflect_apply(Document_querySelector, Window_document_get(), ['body']).textContent;
```

This is an issue, and will likely require rethinking the strategy of only recursing with the current method.

PR #70, opened 2019-09-06
Add type information to created instances

### LINKED ISSUE #35: Handle methods called on created instances
In order to test certain methods for #34 I attempted to add a test that looks like this:

```js
const d = new Document();
const elt = d.createElement('div');
elt.remove();
```

but the `nodeInIDLSpecs` function failed, because `.getType()` returned a strange error value and `.getSymbol()` on that value produced `undefined`. This leaves the above case unhandled.

jackbsteinberg:
Resolves #35


jackbsteinberg:
Created #71 and #72 to capture potential shortcomings of the strategy here.

lib/rewriter.js
```diff
@@ -46,6 +47,13 @@ function rewriter(fileInfo) {
   sourceFile.forEachDescendant((node) => replaceGetSets(node, imports));
   addAllImports(sourceFile, imports);
 
+  // remove JSDocs
```
domenic: Can you also add the TODO pointer to #71 here?

jackbsteinberg: Sure

lib/__testfixtures__/created-instances.input.mjs
```diff
@@ -0,0 +1,8 @@
+const d = new Document();
```
fergald: It's really unclear what is being tested by this. The file is named "created-instances" . I would expect that to have maybe

```js
(new Document()).createElement('div');
(new Document()).location;
(new Document()).length;

const d = new Document()
d.createElement('div');
d.location;
d.length // probably doesn't exist but some test of data property on created instance
```

I don't get why we then go on to make various accesses on `elt.attributes`.

Please add comments to explain what is the purpose of these other lines (in the context of created instances).

Also, perhaps you don't want the first set of things that don't assign to a variable. In that case maybe this file should use "assigned" rather than "created" in its name.

lib/rewriter.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 
+const nodeTypes = require('ts-morph').TypeGuards;
```
fergald: Call this typeGuards?

PR #78, opened 2019-09-12
Mods check prototype chain for true property source

jackbsteinberg:


lib/originals-helper.js
```diff
@@ -56,6 +56,75 @@ function getModuleNameForType(type) {
...
+}
+
+/**
+ * @param {Type} type - a ts-morph type
+ * @param {string} property - the property to find on some module
+ *
+ * @return {string} the get-originals module name corresponding to the
+ * type. (This is sometimes different from the TypeScript type name.)
+ */
+function getModuleNameForTypeFromPrototypeChain(type, property) {
```
domenic: I don't think we need to factor out a separate Type function instead of a Node function. I did that for the other one because we needed to use both at different places in the code. But you only export the Node version, so probably we should just have one.

lib/originals-helper.js
```diff
@@ -56,6 +56,75 @@ function getModuleNameForType(type) {
...
 }
 
+/**
+ * @param {Node} node - a ts-morph AST node
+ * @param {string} property - the property to find on some module
+ *
+ * @return {string} the get-originals module name corresponding to the node's
+ * type. (This is sometimes different from the TypeScript type name.)
+ */
+function getModuleNameForNodeFromPrototypeChain(node, property) {
```
domenic: The two functions `getModuleNameForNode` and `getModuleNameForNodeFromPrototypeChain` behave very differently and have rather different usese, despite their similar names. I think we should differentiate them better in name, and in the jsdoc comments.

One idea is `getModuleNameForPropertyOfNode(node, property)` (or maybe even `(property, node)`?).

The jsdoc should explain that it's not the module name corresponding to the Node's type that you're looking for. It's the module name where that property of the node is found.

lib/originals-helper.js
```diff
@@ -56,6 +56,75 @@ function getModuleNameForType(type) {
...
+  return getModuleNameForTypeFromPrototypeChain(node.getType(), property);
+}
+
+/**
+ * @param {Type} type - a ts-morph type
+ * @param {string} property - the property to find on some module
+ *
+ * @return {string} whether the property exists on the given type
+ */
+function checkPropertyOnType(type, property) {
```
domenic: I'd name this "is" instead of "check"

lib/originals-helper.js
```diff
@@ -56,6 +56,75 @@ function getModuleNameForType(type) {
...
+  for (const symbol of type.getProperties()) {
```
domenic: Can you just do

```js
return type.getProperties().some(symbol => symbol.getName() === property);
```

?

lib/originals-helper.js
```diff
@@ -56,6 +56,75 @@ function getModuleNameForType(type) {
...
+    }
+  }
+
+  // Find the deepest reference to the property in the prototype chain
+  let inspectType = type;
+  let latestFound;
+  while (inspectType !== undefined) {
+    if (checkPropertyOnType(inspectType, property)) {
+      latestFound = inspectType;
+    } else if (latestFound !== inspectType) {
```
domenic: Would it be possible to remove this else if? It seems like line 120 should cover it?

jackbsteinberg: This is an optimization that will break you out of the prototype chain once you've found the most "super" location. I could leave the return to line 120 but I figured if I could add this break I should.

I'm also noticing now that this will fail if the property does not exist on `type`, since latest found will be `undefined`, but that's a case that will only happen if you write code that accesses a property on a global that doesn't exist, which is an error anyway.

lib/originals-helper.js
jackbsteinberg: Yeah good call

lib/__testfixtures__/object-prototype.input.mjs
```diff
@@ -0,0 +1,23 @@
...
+const url = new URL();
+url.toString();
+const bool = new Boolean();
+bool.toString();
+const te = new TypeError();
+te.toString();
+
+// Don't source from Object prototype
+const n = new Number();
+n.valueOf();
```
domenic: Can you test `const n2 = 5; n2.valueOf()` and also just `5..valueOf()`?

lib/__testfixtures__/object-prototype.input.mjs
```diff
@@ -0,0 +1,23 @@
...
+const d = new Date();
+d.valueOf();
+const arr = new Array();
```
domenic: Can you also test `const arr = [5]`?

lib/__testfixtures__/object-prototype.input.mjs
```diff
@@ -0,0 +1,23 @@
...
+elt.toString();
+elt.valueOf();
+elt.hasOwnProperty('remove');
+
+// Should not source from Object but TypeScript is behind
+// https://github.com/jackbsteinberg/get-originals-rewriter/issues/79
+const url = new URL();
+url.toString();
+const bool = new Boolean();
+bool.toString();
```
domenic: Can you also test `const bool2 = false` and just `false.toString()`?

lib/originals-helper.js
```diff
@@ -56,6 +56,61 @@ function getModuleNameForType(type) {
...
+ * @param {Node} node - a ts-morph AST node
+ * @param {string} property - the property for which to find the module
+ *
+ * @return {string} the get-originals module name corresponding to the
+ * type. (This is sometimes different from the TypeScript type name.)
+ */
+function getModuleNameForPropertyOfNode(node, property) {
+  const type = node.getType();
+
+  // Need to use BaseTypeOfLiteralType to catch the type of primitives
```
domenic:
```suggestion
  // Need to use getBaseTypeOfLiteralType to catch the type of primitives
```

lib/originals-helper.js
```diff
@@ -56,6 +56,61 @@ function getModuleNameForType(type) {
...
+  let latestFound;
+  while (inspectType !== undefined) {
+    if (inspectType.getProperty(property) !== undefined) {
+      latestFound = inspectType;
+    } else {
+      break;
+    }
+
+    inspectType = inspectType.getSymbol()
+        .getDeclarations()[0].getBaseTypes()[0];
```
domenic: Add a comment explaining that the 0th base type is the real prototype chain base type, and others are mixins. Maybe with an example.

lib/__testfixtures__/object-prototype.input.mjs
```diff
@@ -0,0 +1,23 @@
...
+const te = new TypeError();
+te.toString();
+
+// Don't source from Object prototype
```
domenic: Can we add tests for strings? In the three variants: `const s = new String()`, `const s2 = ""`, and just `"".valueOf()` directly. (Either valueOf or toString will work.)

Just like numbers, there are some subtleties here around literals vs. the wrapper `String` object type and so on.

lib/__testfixtures__/object-prototype.input.mjs
jackbsteinberg: Yes, I'm pretty sure these tests will fail, since the check for primitives goes before everything else.

PR #84, opened 2019-09-13
Create special case for global method gets and sets

### LINKED ISSUE #57: Getting/setting built-in methods should be rewritten or disallowed
Seen in #56.

```js
const a = new Array();
const globalFunc = a.concat;
```

currently gets rewritten to

```js
import Array, { concat_get as Array_concat_get } from "std:global/Array";
const a = new Array();
const globalFunc = Reflect_apply(Array_concat_get, a);
```

but should be rewritten to

```js
import Array, { concat as Array_concat } from "std:global/Array";
const a = new Array();
const globalFunc = Array_concat;
```

The relevant criteria here is that the property `concat` is a function, and thus a method, not a getter/setter. I think we should basically special-case things that look like get/sets but whose type is Function.

(Technically it's possible to have a getter/setter that returns/accepts a function value, but I don't know of any instances of that across all the specs, and I doubt TypeScript has the type information to distinguish such cases. So, we'd probably need to handle them with a hard-coded list anyway, if such cases do exist.)

---

Similar, but different: consider

```js
const a = [];
a.concat = 5;
```

This should not be rewritten at all. This code (which is bad code) creates a data property on `a`; it does not try to set or override `Array.prototype.concat`. We should let it do that, I guess. Or, we could just disallow it; whichever is easier.

jackbsteinberg:
Resolves #57

### DIFF
--- a/lib/__testfixtures__/function-access.input.mjs
+++ b/lib/__testfixtures__/function-access.input.mjs
@@ -1,2 +1,4 @@
 const a = new Array();
 const globalFunc = a.concat;
+a.concat = 5;
+const other = a.concat;
--- a/lib/__testfixtures__/function-access.output.mjs
+++ b/lib/__testfixtures__/function-access.output.mjs
@@ -1,3 +1,5 @@
 import Array, { concat as Array_concat } from "std:global/Array";
 const a = new Array();
 const globalFunc = Array_concat;
+a.concat = 5;
+const other = a.concat;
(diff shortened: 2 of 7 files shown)


jackbsteinberg:
Currently this specially treats function gets and ignores function sets. However, I noticed in the fringe case alluded to in #57:

```js
a.concat = 5;
const c = a.concat;
```

Will rewrite to

```js
a.concat = 5;
const c = Array_concat;
```

This is currently captured in the failing test. The problem here is that even when put into the ts-ast-viewer `const c = a.concat` shows up as a reference to the global concat method on Array. Should we consider this a case we don't cover and make a note not to do this, or should we write something a more complicated check to try and handle the set -> get case?

jackbsteinberg:
Closing due to inactivity

